PR #42, opened 2025-09-16
shared elements

spencerc99:
https://github.com/user-attachments/assets/5f20b7b4-0d53-43e6-add1-754afba52e12

- after merge need to deploy server


spencerc99:
the room should be derived from the actual room that it uses not just the domain

spencerc99:
@claude what is the inefficient polling you are talking about? i thought we are using websocket-based changes

PR #86, opened 2026-04-14
Handle event API uniformity

spencerc99:
Add dispatchEvent/onEvent to PresenceRoom and playhtml for cleaner event API

- `room.dispatchEvent(type, payload?)` and `room.onEvent(type, callback)` for domain-scoped events via PresenceRoom
- `playhtml.dispatchEvent(type, payload?)` and `playhtml.onEvent(type, callback)` for page-scoped events
- `onEvent` returns an unsubscribe function directly (no ID tracking needed)
- Deprecate `dispatchPlayEvent`, `registerPlayEventListener`, `removePlayEventListener` (still functional, no breaking change)

### DIFF
--- a/.changeset/presence-room-events.md
+++ b/.changeset/presence-room-events.md
@@ -1,0 +1,12 @@
+---
+"playhtml": minor
+"@playhtml/common": minor
+"@playhtml/react": minor
+---
+
+Add dispatchEvent/onEvent to PresenceRoom and playhtml for cleaner event API
+
+- `room.dispatchEvent(type, payload?)` and `room.onEvent(type, callback)` for domain-scoped events via PresenceRoom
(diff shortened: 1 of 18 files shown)


spencerc99:
## Code review

- **P2 [`packages/common/src/index.ts:257`](https://github.com/spencerc99/playhtml/blob/d22560748e97b475a73c950fa8f79c12d06ad603/packages/common/src/index.ts#L257) adds public event APIs without the required docs/starter coverage.** The repo instructions require package behavior/API changes to update relevant `apps/docs/` docs, and require both starter templates to be updated for public core API changes. This PR adds `PresenceRoom.dispatchEvent` / `onEvent` and page-level event APIs, but its changed-file list has no `apps/docs/` files and no `templates/react-starter/` or `templates/html-starter/` files. Add the user-facing docs and starter updates so the public API change is documented where users actually copy from.

Issue #103, opened 2026-04-25
PlayProvider: rethink init-owning vs context-only mode (architecture options)


spencerc99:
## Context

Discovered while debugging a bug where `window.cursorClient` was null on Spencer's website after upgrading to the post-#102 `@playhtml/react`. Root cause turned out to be version skew between `@playhtml/react` and `playhtml` core (one had the new lifecycle, the other didn't), but the investigation surfaced architectural questions worth discussing about the new two-mode `<PlayProvider>` design.

## Current design (Option A): init-owning vs context-only

Post-#102, `<PlayProvider>` has two modes determined by whether `initOptions` is passed:

- `<PlayProvider initOptions={...}>` → init-owning, calls `playhtml.init()`.
- `<PlayProvider>` (bare) → context-only, does **not** call init.

**Pros:**
- Explicit single init site (in theory).
- Aligns with the philosophy that init is a global, idempotent operation.

**Cons:**
- **Breaking change**: previously, `<PlayProvider>` (bare) called `init({})`. Users relying on that get a silently-broken app — context-only providers sit in `isLoading=true` forever if nothing ever calls init.
- **Hydration-order footgun in multi-root apps (Astro `client:only`, micro-frontends)**: if a context-only island hydrates before the init-owning root, it observes `isLoading=true` until the root catches up. Subtle and hard to debug.
- **Implicit mode-switching**: the same component does totally different things depending on whether a prop is passed. The intent isn't grep-able from a usage site.

## Option B: Same options everywhere

Encourage every `<PlayProvider>` to pass the same `initOptions`. `playhtml.init()` is idempotent so duplicates are no-ops.

**Pros:**
- No "which one is the root?" question. Any provider that mounts can trigger init.
- Backward-compatible for the common single-root case.
- Robust to refactors — moving a component into a new island doesn't break it.

**Cons:**
- Configuration duplication / drift risk.
- "Late init-owning provider with different options" silently has its options dropped (current behavior logs a one-time warning, which helps).

## Option C: Construction/provision split (matches React Query / Apollo)

Separate "init the singleton" from "provide context to React":

\`\`\`ts
// Module load or top-level script:
playhtml.init({ cursors: { enabled: true, ... } });

// Anywhere in the React tree(s):
<PlayProvider>
  <App />
</PlayProvider>
\`\`\`

[...]

PR #112, opened 2026-05-03
playprovider init for multiple providers

spencerc99:
Addresses #103.

## Summary

This keeps the existing `<PlayProvider>` bootstrap contract while moving readiness ownership into the core `playhtml` singleton.

- Adds public `playhtml.isLoading` and `playhtml.ready` lifecycle signals.
- Makes duplicate/concurrent `playhtml.init()` calls share the same readiness promise instead of returning early.
- Adopts an existing global `window.playhtml.ready` promise when another compatible playhtml instance has already initialized the page.
- Rejects all waiting init callers when core setup fails instead of leaving duplicate callers pending forever.
- Updates `<PlayProvider>` to still call `playhtml.init()` on mount, but wait for `playhtml.ready` before marking React context as ready.
- Surfaces PlayProvider init failures to React error boundaries instead of marking the provider ready after a failed init.
- Adds `playhtml` as an explicit peer dependency for `@playhtml/react`, now requiring a core version with the readiness contract.

## Why

The architectural question in #103 came from the post-#102 split between init-owning and context-only provider modes. Making bare `<PlayProvider>` context-only was clean conceptually, but it broke existing usage where a bare provider bootstraps playhtml.

This PR takes the lower-risk path:

- Keep bare `<PlayProvider>` backward-compatible.
- Keep configured `<PlayProvider initOptions={...}>` working as before.
- Make core init idempotence stronger so multiple providers and multi-root apps do not race local loading state against global singleton readiness.
- Avoid hangs when React and core observe an already-initialized compatible singleton or a failed setup path.
- Avoid hiding hard init failures behind a falsely-ready React context.

## Implementation Notes

Core now tracks a one-time readiness lifecycle:

- `initStarted` flips synchronously when the first `playhtml.init()` begins.
- `readyPromise` is shared by all duplicate `init()` callers.
- `isLoading` flips false only after the first provider sync path sets up elements and marks them ready.
- Existing compatible `window.playhtml.ready` promises are reused rather than shadowed by an unresolved local promise.
- Setup failures reject the shared readiness promise so callers can fail instead of hanging.
- `resetPlayHTML()` resets the readiness state for test isolation, including cases where setup started but never completed.

React now initializes `hasSynced` from `playhtml.isLoading` and waits on `playhtml.ready` when available. It falls back to the `init()` promise for compatibility with older core shapes during local test/mocking scenarios. If readiness rejects, `<PlayProvider>` stores that failure and throws during render so an app-level error boundary can handle it.

## Validation

- `packages/playhtml`: `/Users/spencerchang/.bun/bin/bun --bun run test` — 20 files, 162 tests passed.
- `packages/react`: `/Users/spencerchang/.bun/bin/bun --bun run test` — 5 files, 23 tests passed.
- `packages/playhtml`: `/Users/spencerchang/.bun/bin/bun --bun run build` — passed.
- `packages/react`: `/Users/spencerchang/.bun/bin/bun --bun run build` — passed.

Builds still emit the existing API Extractor warning that the bundled TypeScript version is older than the project TypeScript version.


spencerc99:
@claude review

PR #116, opened 2026-05-07
cursors: anchor to content when container has a CSS transform

spencerc99:
## Summary

Cursors now anchor to content when the cursor `container` has its own CSS transform (pannable / zoomable canvas). The library reads the live transform matrix from `getComputedStyle()` and stores cursor coordinates in the container's local coordinate space, so two clients with different pan/zoom agree on a cursor's content position; each viewer's CSS transform then maps that position to their own viewport pixels.

- New behavior is opt-in via the existing `cursors.container` option — no new API surface.
- Default (`container = document.body`, no transform) is unchanged.
- `website/fridge.tsx` opts in by setting `container: ".content"`. Cursors now follow the word a user is hovering, regardless of either user's pinch-zoom or pan.
- Updated `CLAUDE.md` to require a changeset for any change under `packages/`.

The math undoes only the matrix's linear (a, b, c, d) part — `getBoundingClientRect()` already absorbs the translate component, so applying the full inverse would subtract translate twice. This was the bug fixed after initial Playwright testing.

## Test plan

- [x] `bun run -C packages/playhtml test` — 156 tests pass, including 2 new tests for the transformed-container path (DOMMatrixReadOnly is stubbed because jsdom doesn't ship one).
- [x] `bunx tsc --noEmit -p packages/playhtml` — clean.
- [x] Two real browser tabs against staging partykit, distinct identities, different pan/zoom transforms. With pageA hovering its "world" word, pageB renders pageA's cursor inside pageB's "world" word's bounding box. Verified with both identity transforms (sub-pixel match) and very different transforms (scale 1.5+pan vs scale 0.8+opposite-pan).
- [x] Manually verified in fridge with two windows.
- [x] Spot-checked non-fridge page — `container` undefined → matrix branch skipped → original behavior preserved.

### DIFF
--- a/.changeset/cursor-transformed-container.md
+++ b/.changeset/cursor-transformed-container.md
@@ -1,0 +1,5 @@
+---
+"playhtml": patch
+---
+
+Cursors now anchor to content when the cursor `container` has its own CSS transform (e.g. a pannable, zoomable canvas). The library reads the live transform matrix from `getComputedStyle()` and stores cursor coordinates in the container's local coordinate space, so two clients with different pan/zoom agree on a cursor's content position; each viewer's CSS transform then maps that position to their own viewport pixels. Default behavior is unchanged when `container` is `document.body` (no transform → identity matrix).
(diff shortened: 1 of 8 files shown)


spencerc99:
@claude review

PR #120, opened 2026-05-09
phase 1b: remove extension/worker (moved to private repo)

spencerc99:
## Summary

Removes `extension/worker/` from this public repo. The Cloudflare Worker now lives in the private repo [`spencerc99/we-were-online-worker`](https://github.com/spencerc99/we-were-online-worker), pinning `@playhtml/extension-types@^0.4.0` for the shared event contract + URL canonicalization helpers.

This finishes the Phase 1 split started in #101 and #117.

### What changes

- **Removed:** `extension/worker/` (every file).
- **Workspaces:** root `package.json` no longer lists `extension/worker`.
- **Changesets:** `.changeset/config.json` no longer ignores `@playhtml/extension-worker` (the workspace no longer exists, so the entry would be unresolved).
- **Lockfile:** `bun.lock` regenerated; one worker-only dep dropped.
- **Code touch-ups:** `extension/src/storage/sync.ts` had two breadcrumb comments referring to the old `extension/worker` path; updated to point at the new private repo.
- **Docs:** `extension/CLAUDE.md` "Worker Backend" section + root `CLAUDE.md` project overview point at the private repo. Worker secrets/setup details stay (still relevant for whoever runs the private repo).

### What does NOT change

- Zero behavior change for users. The deployed worker is the same Cloudflare worker (`playhtml-game-api`); only the deployment source moved.
- Extension client (`extension/src/`) and `wewere.online` experiments call the same endpoints at the same URLs.
- `@playhtml/extension-types` continues to be the single source of truth for the contract.

### Verification

The new private repo was built and verified before this PR was opened:

- Worker source in `we-were-online-worker` builds clean (`bunx tsc --noEmit` only shows the 4 pre-existing `export.ts` warnings that were on main before).
- `bun run test` passes 18/18.
- `wrangler dev` starts locally, `/events/recent` returns real data from Supabase.
- `wrangler deploy --dry-run` produces a valid 2.4 MiB → 473 KiB gzipped bundle.
- Production secrets (`SUPABASE_URL`, `SUPABASE_SECRET_KEY`, `ADMIN_KEY`, `RESEND_API_KEY`) confirmed set on the existing Cloudflare worker.

Public-repo verification:

- [x] `bun build-packages` succeeds.
- [x] `cd extension && bun run build` succeeds.
- [x] `bun run build-site` succeeds.
- [x] `cd extension && bunx vitest run` — 117/117 pass.

### After merge

The next worker deploy must come from the private repo (`bun run deploy` in `we-were-online-worker`), NOT from this PR's commit. Deploys triggered from this repo will fail at build time because the worker source is gone.

Spec: `internal-docs/specs/2026-04-23-worker-split-and-experiment-gating.md` — Phase 1b.


spencerc99:
Closing without merging.

Reorienting on the worker-privacy question: the visualization-changes-need-worker-changes friction from a two-repo split outweighs the benefit of source privacy, and the actual scrape protection comes from runtime auth (signed tokens) and edge defenses (rate limits, caching, WAF) rather than from source visibility.

Next: bring the runtime defenses into the public repo. New plan in flight at `internal-docs/plans/2026-05-10-worker-auth-and-hardening.md`.

The private repo `spencerc99/we-were-online-worker` will be deleted since it has no commits unique to it.

PR #156, opened 2026-06-01
Reduce Cloudflare cost from extension-owned rooms

spencerc99:
## Summary

- Gate extension-owned PlayHTML startup so it only runs for explicitly supported cursor sites when no native PlayHTML instance is present.
- Keep high-traffic sites like YouTube, Google, X, Reddit, and TikTok disabled in the extension policy, while retaining Wikipedia support with stricter hostname matching.
- Catch known Durable Object overload errors at the Worker boundary and return `503 Service Busy` with `Retry-After` instead of letting overloads escape as Worker exceptions.
- Remove routine PartyKit connection, successful epoch-check, normal autosave, and bridge-apply logs while keeping warnings/errors and compaction/reset signal.

## Rationale

The Cloudflare cost spike was driven by extension-created rooms on high-traffic pages, especially YouTube `/watch`, plus large volumes of low-signal Worker/Durable Object logs. The extension fix prevents new supported builds from initializing PlayHTML on those pages, and the PartyKit overload handling reduces noisy exception logging when existing/old clients still hit overloaded Durable Objects.

## Validation

- `bun run test run src/__tests__/presencePolicy.test.ts`
- `bun test partykit/__tests__/serverLimits.test.ts`
- `bunx tsc -p partykit`
- `git diff --check origin/main...HEAD`

## Code review

Requested a pre-merge code review using the `superpowers:requesting-code-review` workflow. Result: no Critical or Important issues. Minor non-blocking suggestion: add a future boundary-level test for the Worker overload `503 Service Busy` response.

## Notes

This branch was recreated from current `origin/main` and cherry-picked to keep the PR scoped to the Cloudflare cost/logging fixes only.

### DIFF
--- a/partykit/party.ts
+++ b/partykit/party.ts
@@ -1206,6 +1206,11 @@ export class PartyServer extends YServer {
       );
     }
 
+    // Log structured information about the save
+    console.log(
+      `[PartyServer] Autosave: room=${this.name}, size=${documentSize} bytes (${(documentSize / 1024 / 1024).toFixed(2)} MB), resetEpoch=${docResetEpoch ?? serverResetEpoch ?? "none"}`
+    );
+
     // Save the document to the database
     const { data: _data, error } = await supabase.from("documents").upsert(
       {
--- a/partykit/party.ts
+++ b/partykit/party.ts
@@ -706,6 +706,10 @@ export class PartyServer extends YServer {
     const compactAfter = Date.now() + DEFAULT_EMPTY_ROOM_COMPACT_DELAY_MS;
     await this.setEmptyRoomCompactAfter(compactAfter);
     await this.scheduleNextAlarm();
+
+    console.log(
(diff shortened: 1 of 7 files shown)


partykit/party.ts
```diff
@@ -1216,11 +1206,6 @@ export class PartyServer extends YServer {
       );
     }
 
-    // Log structured information about the save
-    console.log(
-      `[PartyServer] Autosave: room=${this.name}, size=${documentSize} bytes (${(documentSize / 1024 / 1024).toFixed(2)} MB), resetEpoch=${docResetEpoch ?? serverResetEpoch ?? "none"}`
-    );
-
```
spencerc99, line 1223: can we keep this?

spencerc99: Restored this autosave size/resetEpoch log in bcfdede. I kept it as the attempt log before the database write and left the separate autosave-success and bridge-apply logs removed.

PR #165, opened 2026-06-06
load-test: walking-together scenario (100-participant cursor + URL load)

spencerc99:
## Summary

Adds a `walking-together` scenario to the load-test harness so we can verify the page holds up with ~100 active participants — the scenario that crashed the room before the keyed-map roster fix.

Each virtual user behaves like a real participant: continuous cursor movement (awareness, 10Hz), periodic keyed-map roster upsert, and periodic URL shares to the chat array.

Also fixes the stale root `package.json` `load-test` script paths (`load-test/` → `tools/load-test/`, where the harness was moved in an earlier refactor but the scripts weren't updated).

## Results (local wrangler dev, 100 users, transient mode)

- **0 client errors, 100% propagation** — the room stayed up the whole run.
- **No DO overload / 503 / memory failure** — verified in the worker log (the earlier "503" count was a false grep match on a hex connection id).
- Latency degrades (p95) past ~75 users, but that's a single local DO on a laptop, not production infra — it's degradation, not collapse.
- The one real server-side event: the **per-connection message-rate limit (1000 msg/sec)** closed some very-chatty synthetic clients, which auto-reconnect. Worth knowing for very active real workshops, but not a failure.

## How to run

```bash
# Against a local wrangler dev worker (recommended — never load-test prod):
SUPABASE_URL=http://127.0.0.1:9 SUPABASE_KEY=bad ADMIN_TOKEN=dev \
  bunx wrangler dev --config partykit/wrangler.jsonc --port 1999 --var SUPABASE_LOAD_TIMEOUT_MS:200 &
bun tools/load-test/src/runner.ts --scenario walking-together --users 100 --duration 30 --target local
```

Doc-bounded-ness is covered by the companion soak test (`smoke:partykit:soak`), which measures struct count from a connected client (the load-test admin-inspect can't read live doc size in transient mode).


spencerc99:
## Code review

PR #175, opened 2026-06-12
Add report-only extension performance tracing workflow

spencerc99:
## Summary

- adds `perf:extension:*` package scripts for running the trace harness, comparing summaries, and testing comparison logic
- adds a report-only GitHub Actions workflow that builds base and head workspace packages plus extension bundles, traces both under Chromium, writes a step-summary report, and uploads trace/report artifacts
- adds a pure comparison script that flags large TaskDuration/ScriptDuration/Layout/RecalcStyle/heap regressions while exiting successfully unless `--fail-on-regression` is passed
- shows both relative and absolute metric deltas in the Markdown report so flagged changes are readable without opening JSON artifacts
- defaults the trace harness to Playwright Chromium when local macOS Chrome paths are not available, so CI can run it under Linux
- keeps the heavier local trace scenario available while letting CI pass smaller `--mouse-steps`/`--scroll-steps` values that still exercise cursor, keyboard, and scroll collection without timing out on hosted runners
- removes temporary Chromium profile directories after each trace run so uploaded artifacts stay focused on traces and reports

## Validation

- stacked on rebased #173 head (`b7ec97e`), which is based on latest `main` (`518c8d0`)
- `bun run perf:extension:test`: 2 tests passed
- `node --check scripts/compare-extension-performance.mjs`: passed
- `node --check scripts/trace-extension-performance.mjs`: passed
- `git diff --check`: passed
- `bun run test -- run` from `extension/`: 35 files, 239 tests passed
- `bun build-packages`: passed with existing API Extractor TypeScript-version warnings
- `bun run build-extension`: passed
- `bun run build:firefox` from `extension/`: passed
- sample compare CLI run against a temp trace summary: passed and wrote Markdown/JSON artifacts
- GitHub Actions `pkg.pr.new` run #387: passed
- GitHub Actions `Extension Performance Report` run #8: passed end to end

## Notes

This PR intentionally keeps the workflow report-only. It flags dramatic metric increases in the report and artifacts, but does not fail PRs until we have enough trace history to tune thresholds.

The first workflow run exposed that clean CI needs `bun build-packages` before `bun run build-extension`; the workflow now mirrors that dependency build order. Later runs showed the full local cursor movement scenario is too heavy for hosted CI under tracing, so the workflow uses a smaller automated scenario while the local trace script still defaults to the heavier run.


spencerc99:
## Code review

PR #188, opened 2026-06-18
Fix React shared id precedence

### LINKED ISSUE #38: bug: inconsistent behavior with the `id` prop on react package
when the `id` prop passed into `withSharedState` conflicts with the topline element id prop, the behavior is confusing. Also seems to inconsistently use the prop, like sometimes doesn't work. 
- should show a warning (maybe a linter warning?) if there is a conflict
- should probably have the prop directly passed in to `withSharedState` take precedence over the one that is on the main element

spencerc99:
## Summary

Fixes the confusing `id` precedence behavior in `@playhtml/react` when a `withSharedState` configured id conflicts with the rendered child element id.

- Applies the configured `id` to the rendered DOM element so `withSharedState` takes precedence over a child id.
- Warns when the configured id and child id conflict.
- Cleans up the previously registered binding before registering a changed configured id, so React DOM node reuse does not leave stale handlers behind.
- Reports `dataSource` precedence accurately in conflict diagnostics.
- Warns on explicit empty configured ids and dedupes conflict warnings across rerenders.
- Adds a patch changeset for `@playhtml/react`.

## Root Cause

`withSharedState` passed `id` into `CanPlayElement`, but normal DOM children kept their own id because the configured id was only used for fragments. `setData` resolves handlers from the actual effective DOM id, so conflicting ids could bind to the child id instead of the configured id. The first fix exposed lifecycle and diagnostic edge cases around changed ids, `dataSource`, empty ids, and repeated warnings.

Closes #38.

## Verification

- `bun run test src/__tests__/elements.test.tsx` -> 15 tests passed
- `bun run test` -> 5 files passed, 34 tests passed
- `bun run build` -> passed `tsc && vite build` with the existing API Extractor TypeScript-version warning
- `git diff --check` -> clean
- Omegacode review was run; follow-up findings were patched in `352f39e`


spencerc99:
## Code review

- **P2 [`packages/react/src/index.tsx:461`](https://github.com/spencerc99/playhtml/blob/b97e06fa7e5c4972be4eca0c4cf5cfb090c37281/packages/react/src/index.tsx#L461) changes React shared-state `id` behavior without the required docs-audit handling.** The repo instructions require package behavior/API changes to update relevant `apps/docs/` docs when they affect how users write React components, and if no docs change is needed the PR summary must say why. This PR makes `withSharedState` apply the configured `id` to the rendered DOM element and changes conflict diagnostics, but the changed-file list has no `apps/docs/` files and the PR body does not state why docs are unnecessary. Either add the relevant React API guidance/docs update or explicitly document why no public docs change is needed.

PR #190, opened 2026-06-18
Reactive `view` API for vanilla can-play (RFC #95)

### LINKED ISSUE #95: RFC: Reactive view API for vanilla `can-play` (proposed design for feedback)
We're planning a refactor of playhtml's vanilla authoring API to give `can-play` proper reactive view updates (today anything complex forces authors into `innerHTML` replacement or hand-rolled DOM mutation). This RFC proposes a concrete API and is here to gather feedback on its **ergonomics** before we build it.

If you've written (or tried to write) custom `can-play` capabilities, your reactions here will directly shape what we ship.

> **Edit (v2):** this RFC originally proposed two parallel tracks (`can-mirror` composition upgrades + `can-play` views). Based on feedback it's now narrowed to **`can-play` + `view` only**, with a tightened API contract, corrected examples, and a rendering-lifecycle walkthrough. The `can-mirror` work is split into a follow-up (see "Related rail" at the bottom).

## Context

playhtml gives HTML magic interactive powers via data attributes — `can-move` for drag, `can-toggle` for click-to-toggle, `can-spin` to rotate, and so on. The library's ethos is "write normal HTML, sprinkle an attribute, and the element becomes collaborative." No framework, no build step, no component model.

This works great for the built-in capabilities, but not for custom ones. **`can-play`** lets you define your own data model and update logic, but `updateElement` only hands you data and says "go mutate the DOM" — there's no reactive model. The actual `can-post` guestbook on playhtml.fun is 80+ lines that have to:

- Track which entries were already rendered (a local `Set`) to avoid re-rendering everything on every update
- Hand-build DOM with `createElement` + `innerHTML` for structure, then `querySelector` + `textContent` to safely inject user text (XSS workaround)
- Manually handle insert-at-top semantics with `prepend`
- Punt entirely on removals, reorders, and edits

The workaround today is React (`@playhtml/react`), which defeats the point of playhtml being "collaborative HTML for people who know HTML." This RFC proposes the vanilla path.

## The proposal in one sentence

New `playhtml.register(elementId, init)` / `playhtml.define(tagName, init)` APIs and a new optional `view` field on the capability definition: `view` is a pure function from state to a [lit-html](https://lit.dev/docs/libraries/standalone-templates/) template, and playhtml patches the element to match it whenever state changes. `register` binds one element; `define` registers a reusable capability for any element carrying the attribute.

`view` is **sugar over `updateElement`**: same lifecycle, same triggers — but instead of mutating the DOM yourself, you return a description of it and lit-html does the diffing.

## How rendering works

The HTML element is a mount point — an empty shell:

```html
<div id="counter" can-play></div>
```

1. **Register.** `playhtml.register("counter", { defaultData, view })` puts the definition in a registry. Nothing renders yet.
2. **Bind.** When both the definition and the element exist (in either order — same model as `customElements.define`), playhtml creates the element handler, exactly like every built-in capability.
3. **First render.** Where the handler would call `updateElement` on mount, it instead calls lit-html's `render(view(ctx), element)`, writing the template's DOM into the div.
4. **Interact.** `@click=${...}` in the template is a real event listener. It calls `setData` — the existing write path, nothing new.
5. **Re-render.** When data changes (locally or from a remote client via Yjs), the same "data changed → update element" trigger that calls `updateElement` today re-invokes `view` and calls `render()` again. lit-html patches only the difference — a changed text node updates in place; buttons aren't recreated, listeners aren't re-attached.

The implementation surface is intentionally small: `view` is a sibling of `updateElement` on `ElementInitializer`, the handler's update pipeline branches `view ? render(view(ctx), el) : updateElement(ctx)`, and lit-html does all the hard reconciliation work.

### `playhtml.register(elementId, init) → PlayElementHandle`

```ts
type PlayElementHandle<TData, TAwareness = unknown> = {
  id: string;
  getElement(): HTMLElement | null;     // null until the element exists and is bound
  getData(): Readonly<TData>;           // read-only snapshot — ALL writes go through setData
  setData(next: TData | ((draft: TData) => void)): void;
  setMyAwareness(next: TAwareness): void;
  unregister(): void;                   // detach handlers, run onMount cleanup; shared data is preserved
};
```

Rules:

- `register()` may be called **before or after** `playhtml.init()`, and before or after the element exists in the DOM. Binding happens once both exist (upgrade semantics, like custom elements). This kills today's config-before-init ordering footgun.
- The element must have a stable, unique `id` (it *is* the `elementId`).
- The `can-play` attribute is **optional** when registering by id — `register()` implies it. Dev builds warn if the id never appears in the DOM after load.
- Re-registering the same id replaces the initializer; persisted data is untouched.
- `onMount` may return a cleanup function, which runs on `unregister()`.
- `handle.getData()` returns a read-only snapshot. Mutating it does nothing (and warns in dev). The only write paths are `setData`'s two forms.
- `playhtml.getHandle(elementId)` returns the same handle for any bound element (useful when the registration happened elsewhere, e.g. for `define`-based elements below).

### `playhtml.define(tagName, init)` — reusable capabilities

`register` binds one element. For "20 sticky notes with the same behavior," `define` registers a reusable **capability** under an attribute name, and every element carrying that attribute gets it:

```html
<div id="note-1" can-note data-color="yellow"></div>
<div id="note-2" can-note data-color="pink"></div>

<script type="module">
  import { playhtml, html } from "https://unpkg.com/playhtml@latest/dist/full.es.js";

  playhtml.define("can-note", {
    defaultData: (element) => ({ text: "", color: element.dataset.color ?? "yellow" }),
    view: ({ data, setData }) => html`
      <textarea
        class="note note--${data.color}"
        .value=${data.text}
        @input=${(e) => setData((draft) => { draft.text = e.target.value; })}
      ></textarea>
    `,
  });

  playhtml.init();
</script>
```

Rules:

- Same registry/upgrade semantics as `register`: callable before or after `init()`, and elements with the attribute are bound whenever they appear — including ones added to the DOM later.
- Each bound element still needs a stable unique `id`; per-element data is keyed by it, exactly like built-in capabilities.
- `register(elementId, init)` is internally just `define` scoped to a single element — one mental model, two entry points.
- Today's `init({ extraCapabilities })` becomes sugar for `define` (one registry); defining a name that collides with a built-in capability is an error, and re-defining a custom name replaces the initializer.

### The definition

```ts
playhtml.register("guestbook", {
  defaultData: [],                       // or (element) => T, as today
  myDefaultAwareness: { typing: false },

  // NEW — same context-object style as updateElement:
  view: ({ data, setData, awareness, setMyAwareness, element }) => html`...`,

  onMount: ({ getElement, setData, setMyAwareness }) => {
    // optional non-DOM setup (intervals, document-level listeners, ...)
    return () => { /* optional cleanup */ };
  },

  resetShortcut: "shiftKey",
});
```

- `view` re-renders on **data and awareness changes**. (`localData` is deliberately left out of the view context for v1 — see open questions.)
- `view` and `updateElement` are **mutually exclusive** — registration-time error, not a silent precedence rule. `onClick`/`onDrag` are likewise disallowed alongside `view` (use `@click` etc. in the template). `onMount` and `resetShortcut` still apply.
- `view` must be pure: calling `setData` synchronously during render is a dev-mode error. (This bakes the write-loop protection from our data rules into the API.)

### Writes

`setData` keeps its existing two forms, unchanged:

```ts
setData({ on: true });                  // value form: replace the snapshot
setData((draft) => { draft.count++ });  // mutator form: merge-friendly CRDT edit
```

One clarification the original RFC got wrong: **the mutator must mutate the draft, not return a new object.** `setData(d => ({ count: d.count + 1 }))` is a silent no-op today (the return value is discarded). We'll add a dev-mode warning when a mutator returns a non-`undefined` value, same as Immer.

For writes from code *outside* the element (form handlers, buttons elsewhere on the page), use the handle:

```ts
guestbook.setData((draft) => { draft.unshift(entry); });
```

### Imports

One module, no global/import mixing:

```html
<script type="module">
  import { playhtml, html, repeat } from "https://unpkg.com/playhtml@latest/dist/full.es.js";
</script>
```

lit-html ships **in the core library** — at ~7 KB gzip it's not worth a split bundle, and it means `view` works everywhere playhtml does. We re-export a curated subset: `html`, `repeat`, `classMap`, `styleMap`. Deliberately **no `unsafeHTML`** — auto-escaping of interpolated values is half the pitch (it deletes the entire `querySelector` + `textContent` XSS dance from the current guestbook). For script-tag users the same helpers hang off the global (`const { html, repeat } = playhtml`).

### Counter

```html
<div id="counter" can-play></div>

<script type="module">
  import { playhtml, html } from "https://unpkg.com/playhtml@latest/dist/full.es.js";

  playhtml.register("counter", {
    defaultData: { count: 0 },
    view: ({ data, setData }) => html`
      <button @click=${() => setData((draft) => { draft.count -= 1; })}>-</button>
      <span>${data.count}</span>
      <button @click=${() => setData((draft) => { draft.count += 1; })}>+</button>
    `,
  });

  playhtml.init();
</script>
```

### Shared lamp — state → style

```html
<div id="lamp" can-play></div>

<script type="module">
  import { playhtml, html } from "https://unpkg.com/playhtml@latest/dist/full.es.js";

  playhtml.register("lamp", {
    defaultData: { on: false },
    view: ({ data, setData }) => html`
      <button
        class=${data.on ? "lamp is-on" : "lamp"}
        aria-pressed=${data.on ? "true" : "false"}
        @click=${() => setData((draft) => { draft.on = !draft.on; })}
      >
        lamp
      </button>
    `,
  });

  playhtml.init();
</script>
```

### Guestbook — rewriting the 80-line pain example

```html
<div id="guestbook" can-play></div>

<form id="guestbook-form">
  <input name="name" required />
  <input name="message" maxlength="140" required />
  <button type="submit">🖋️</button>
</form>

<script type="module">
  import { playhtml, html, repeat } from "https://unpkg.com/playhtml@latest/dist/full.es.js";

  const guestbook = playhtml.register("guestbook", {
    defaultData: [],
    view: ({ data: entries }) => html`
      <ul>
        ${repeat(entries, (e) => e.id, (e) => html`
          <li>
            <strong>${e.name}</strong>
            <span>${e.message}</span>
            <time datetime=${new Date(e.at).toISOString()}>${formatDate(e.at)}</time>
          </li>
        `)}
      </ul>
    `,
  });

  document.getElementById("guestbook-form").addEventListener("submit", (event) => {
    event.preventDefault();
    const form = event.currentTarget;
    const fields = Object.fromEntries(new FormData(form));

    guestbook.setData((draft) => {
      draft.unshift({
        id: crypto.randomUUID(),
        name: String(fields.name),
        message: String(fields.message),
        at: Date.now(),
      });
    });

    form.reset();
  });

  playhtml.init();
</script>
```

No manual diffing, no render-tracking `Set`, no `createElement`/`querySelector` dance, and user text is auto-escaped. `repeat` keyed by a stable unique id handles inserts, removals, and reorders.

## Data modeling notes

- **Key lists by a stable unique id** (`crypto.randomUUID()`), never by timestamp — two users submitting in the same millisecond is exactly the kind of thing playhtml makes routine.
- **For lists where entries get edited or reordered**, a flat array of objects starts to strain; the next step up is a two-part shape — `{ entriesById: {...}, entryOrder: [...] }` — where the order array holds ids and the map holds the records. Views should guard against ids missing from the map (`if (!entry) return null`) since concurrent trims can briefly orphan ids. We'll document this pattern, but the flagship example stays a flat array because append-only doesn't need it.
- **Migrating live elements:** `defaultData` only seeds brand-new rooms. If an element already has persisted data (like the production guestbook, a top-level array whose entries have no `id` field), the rewrite must keep the persisted shape and read defensively (e.g. key with a fallback: ``(e) => e.id ?? `${e.name}-${e.at}` ``) — or use a new element id. A shape change against a live room crashes it.

## Related rail: `can-mirror`

`can-mirror` remains the right mental model when **existing DOM is the source of truth** ("I have HTML, make it collaborative") — versus `can-play` + `view` for "I have data, render a view of it." The composition upgrades previously sketched here — keyed reconciliation, descendant capability boundaries, ignored subtrees, scoped sync — are a separate design with its own hard problems (concurrent-DOM-edit conflict semantics, form state) and will get their own RFC. This one is `can-play` + `view` only.

## What we want feedback on

1. **Does `playhtml.register(elementId, init)` → handle feel right** as the author path? Is the handle the right home for outside-the-element reads/writes?
2. **Does the `register` / `define` split make sense** — one-off element vs. reusable capability — or should it be a single function?
3. **Is lit-html the right template library?** We considered Spellcaster, Alpine.js, petite-vue, and writing our own. lit-html won on: renderer-only (not a framework), actively maintained, shadow-DOM-compatible for future web-components plans, battle-tested keyed list reconciliation, and auto-escaping by default. Note that re-exporting `html`/`repeat` makes lit-html's template semantics part of playhtml's public API, and it ships in the core bundle (~7 KB gzip) — strong opinions welcome.
4. **Should `localData` be in the `view` context in v1**, or is data + awareness enough?
5. **Any missing cases?** Complex capabilities you've wanted to write and couldn't, or hacked around — paste examples, they'll stress-test the design.

spencerc99:
Implements the reactive `view` API proposed in #95, giving vanilla `can-play` a declarative render path so custom capabilities can render from state instead of hand-mutating the DOM.

## What's here

**Core API** (`@playhtml/common` + `playhtml`):
- **`view`** — a new optional field on the capability definition: a pure `state → lit-html template` function that playhtml patches into the element on every data / localData / awareness change. Mutually exclusive with `updateElement` (and with element-level `onClick`/`onDrag`, which move into the template as `@click` etc.), validated at registration.
- **`playhtml.register(elementId, init)`** — binds a `view`/`updateElement` initializer to a single element by id and returns a handle (`getElement`, `getData`, `setData`, `setLocalData`, `setMyAwareness`, `requestUpdate`, `unregister`) for reads/writes from outside the view. Callable before or after `init()` and before or after the element exists (registry/upgrade binding, like `customElements.define`).
- **`playhtml.define(tagName, init)`** — registers a reusable capability under an attribute name; every element carrying it binds, including mount points rendered by a view (post-render descendant binding → composition). The imperative counterpart of `init({ extraCapabilities })`.
- **`playhtml.getHandle(elementId)`** — returns a handle for any bound element.
- **`requestUpdate()`** (handle + event/`onMount` context) re-runs a view without a data change — for clock-driven views (timers, relative-time labels, animations).
- **`setLocalData`** now accepts a mutator function and re-renders in view mode, so per-user UI state lives in `localData`.
- `html`, `svg`, `nothing`, `repeat`, `classMap`, `styleMap` are re-exported from `playhtml` (lit-html ships in core; **no `unsafeHTML`**, so interpolated values stay auto-escaped).

**Guardrails:** `setData` called synchronously during a render is rejected (re-render-loop protection); a `setData` mutator that returns a value warns in development. Existing `updateElement`-based capabilities are unaffected — `view` is purely additive.

**Playground:** `website/test/view-playground.html` (sibling to `/test/react-test`) ports the big `@playhtml/react` examples to the new vanilla API — counter, shared lamp, slider, online indicator, visitor count, timer (`requestUpdate`), random spinner (SVG + `requestUpdate`), poll (keyed votes), live reactions (events + fade), live chat (`define` + `localData` + view-rendered mount points), and a guestbook. Each section is an empty `can-play` mount point; all markup is rendered by its `view`.

## Testing

- New `view-api.test.ts`: 11 tests covering render/re-render, handle reads/writes, awareness re-render, `requestUpdate`, `setLocalData` re-render, the during-render write guard, validation throws, `define` multi-instance, view→mount-point composition, and `svg` fragments.
- Full suite green after rebasing onto current `main`: **playhtml 197**, **react 29**. All core packages typecheck clean.
- Playground HTML verified to transform under vite (bare `playhtml` import resolves; no transform errors). Note: it hasn't been clicked through in a live synced browser — that needs a running partykit room.

## Notes for review

- This is the RFC implementation; per the audit on #95, two follow-ups would most improve ergonomics for the back third of the example catalogue: an **effect-on-data-change hook** (audio/scroll/focus side effects) and **presence/cursor changes auto-re-rendering a view** (element awareness already does). Neither is in this PR.
- Changeset included (`playhtml` + `@playhtml/common`, minor).

Closes #95.


spencerc99:
## Code review

- **P2 [`packages/playhtml/src/index.ts:1585`](https://github.com/spencerc99/playhtml/blob/8580f01397ca5676e237056f99f7e20e2ddfd0e3/packages/playhtml/src/index.ts#L1585) adds public `register` / `define` / `getHandle` APIs without the required starter-template updates.** The repo instructions require both `templates/react-starter/` and `templates/html-starter/` to be updated whenever a public core API changes, including exported functions and vanilla `can-play` API changes. This PR exposes the new view API on `playhtml`, but the changed-file list has no files under either starter template. Add starter examples/usage updates for the new vanilla API and verify both starters as required.

packages/playhtml/src/index.ts
```diff
@@ -1315,7 +1320,22 @@ function createPlayElementData<T extends TagType, TData = any>(
...
+            isDevelopmentMode &&
+            returned !== undefined &&
+            typeof returned === "object"
+          ) {
+            console.warn(
+              `[playhtml] A setData() mutator for "${elementId}" returned an object. ` +
+                `Mutators must mutate the draft in place (e.g. \`d => { d.count++ }\`); ` +
+                `the return value is ignored. To replace the whole snapshot, pass a value instead of a function.`,
+            );
+          }
```
spencerc99, line 1338: why do we have this new guard?

packages/playhtml/src/index.ts
```diff
@@ -1355,10 +1375,12 @@ function isCorrectElementInitializer(
 ): tagInfo is ElementInitializer {
   return (
     tagInfo != null &&
+    // Any non-undefined default is valid — objects, functions, and primitives
+    // (e.g. `defaultData: 0` or `""`), which views commonly use.
```
spencerc99: hmm no we want to enforce that its always an object so that its robust to changing the data

packages/playhtml/src/index.ts
```diff
@@ -2120,10 +2176,303 @@ function removePlayElement(element: Element | null) {
...
+  /** Current shared data. Do not mutate it directly — mutations are not synced and may corrupt state; write via setData. */
+  getData(): T | undefined;
+  setData(next: T | ((draft: T) => void)): void;
+  setLocalData(next: U | ((draft: U) => void)): void;
+  setMyAwareness(next: V): void;
+  /** Re-run the view now (for clock-driven views). No-op without a view. */
+  requestUpdate(): void;
+  /** Detach the handler (shared data is preserved) and drop the registration. */
+  unregister(): void;
+}
```
spencerc99, line 2232: how is this different frm our main element handler in the common package? is there any way we can reuse code?

packages/playhtml/src/index.ts
```diff
@@ -2120,10 +2176,303 @@ function removePlayElement(element: Element | null) {
...
+  }
+  if (!init.view && !init.updateElement) {
+    throw new Error(
+      `[playhtml] "${name}" must define either \`view\` or \`updateElement\`.`,
+    );
+  }
+}
+
+/** Stamps a registration's initializer fields onto its element as props. */
+function stampRegistrationOntoElement(
```
spencerc99, line 2276: hmm i see this is actually better... i wonder if we should just do this kind of path / API for the imperative can-play api too where you do something like playhtml.register(XXX) and don't have to put can-play on the element - bc you can't just put it by itself you need to define other stuff too

spencerc99: i guess the argument was that it's more straightforward / simpler code to set it directly on the element but i feel like it might be more confusing since you dont know what's happening?

packages/playhtml/src/index.ts
```diff
@@ -2120,10 +2176,303 @@ function removePlayElement(element: Element | null) {
...
+      `[playhtml] register("${elementId}") — no element with that id is in the DOM yet. ` +
+        `It will bind automatically when the element appears.`,
+    );
+  }
+  // register always binds via can-play, so scope the handle to that tag.
+  return createPlayElementHandle(
+    elementId,
+    TagType.CanPlay,
+  ) as PlayElementHandle<T, U, V>;
+}
```
spencerc99, line 2394: hmm i wonder if this can overlap with setupPlayElement or ocllapse into the same function? If it's the view API handle the view api otherwise..?

spencerc99: maybe we should leave these simplification changes until we cofnirm this is good to merge? idk it feels like a good path forward regardless having some non-imperative path like this for normal javascript

packages/playhtml/src/index.ts
```diff
@@ -2120,10 +2176,303 @@ function removePlayElement(element: Element | null) {
...
+ * `init({ extraCapabilities })`.
+ *
+ * @param capabilityName - The attribute name elements use to opt in (used in an
+ *   attribute selector, e.g. `[can-note]`).
+ * @experimental New view API; signature may change in a future minor release.
+ */
+function definePlayCapability<T = any, U = any, V = any>(
+  capabilityName: string,
+  init: ElementInitializer<T, U, V>,
+): void {
```
spencerc99, line 2409: same thing with this if we can consolidate with the other logic we have to define a new tag?

packages/playhtml/src/index.ts
```diff
@@ -2120,10 +2176,303 @@ function removePlayElement(element: Element | null) {
...
+ *
+ * Reconciliation, not just binding: newly-present descendants are bound, and
+ * descendants this root previously bound that are now gone (a keyed list item
+ * was removed) are torn down via removePlayElement — which runs their onMount
+ * cleanup and disconnects their observers, so churning lists don't leak
+ * handlers pointing at detached DOM. Shared data is preserved.
+ */
+const viewDescendants = new WeakMap<HTMLElement, Map<string, HTMLElement>>();
+
+function setupViewDescendants(root: HTMLElement): void {
```
spencerc99, line 2452: can any of our mutation observer logic here be combined with our logic for can-mirror? i feel like we are doing a lot of the same things

apps/docs/src/content/docs/custom-elements.mdx
```diff
@@ -0,0 +1,584 @@
...
+    <p class="ph-canplay-inline__caption">
+      Drop a note about what you're learning or building. The list is shared, capped at 20, oldest entries fall off.
+    </p>
+  </div>
+  <DocsGuestbook client:only="react" />
+</div>
+
+### Lists
+
+This is where the view API earns its keep — `repeat(items, keyFn, template)` is lit-html's keyed list, versus hand-diffing DOM in `updateElement`. **Key by a stable unique id** (`crypto.randomUUID()`), never an index or timestamp.
```
spencerc99, line 304: im confused by the layout of the docs here we have lit-html in the copy for everyone but the example shows html and react.. i think the toggle should be further above?

apps/docs/src/content/docs/custom-elements.mdx
```diff
@@ -0,0 +1,584 @@
...
+
+### Dynamically-added elements
+
+`playhtml.init()` scans the DOM once. For elements you add later, call `playhtml.setupPlayElement(el)` after inserting them (and use [`selector-id`](/docs/advanced/dynamic-elements/) for many-of-a-kind elements). See [Dynamic elements](/docs/advanced/dynamic-elements/).
+
+:::note[Lists and data-driven UIs get verbose here]
+With `updateElement` you hand-build and diff the DOM yourself, which is painful for lists, chat, and anything computed from a collection. That's exactly what the [reactive view API](#reactive-view-api-experimental) below is for — and React's `withSharedState` already renders declaratively.
+:::
+
+## can-mirror
```
spencerc99: we should show can-mirror before can-play

We should also just segment the entire doc by vanilla vs react.

Vanilla should go can-mirror -> can-play and at the top share diff between the two: can-mirror syncs everything and puts state in the DOM, can-play gives you fine-grained control over data. 

React only has the withSharedState path and should just focus on that

spencerc99: they should use the same examples and ideally share the same state for the live demos of those examples

apps/docs/src/content/docs/data/data-essentials.md
```diff
@@ -112,15 +112,14 @@ defaultData: {
...
-// Good
+// Good — derive in the view
 defaultData: { count: 5 }
-updateElement: ({ element, data }) => {
-  const parity = data.count % 2 === 0 ? "even" : "odd";
-  element.textContent = `${data.count} (${parity})`;
-}
+view: ({ data }) => html`
+  ${data.count} (${data.count % 2 === 0 ? "even" : "odd"})
+`
```
spencerc99, line 130: can you show both of the examples for vanilla ? Like show either X or Y

packages/playhtml/src/index.ts
spencerc99: It's a dev-only footgun catcher for `setData` mutators: if you write `setData(d => ({ ...d, x: 1 }))` (returning a new object) instead of mutating in place `setData(d => { d.x = 1 })`, the return value is silently ignored and your write is lost. The guard warns on exactly that case — only in dev mode, and only when the return is an object, so terse valid mutators like `d => d.count++` (which return a number as a side effect) stay quiet. No runtime behavior, just a warning.

---

packages/playhtml/src/index.ts
spencerc99: It's not a second handler — `PlayElementHandle` is a thin public *facade* over the one `ElementHandler`. Every method does `findHandlerForElementId(id, tag)` and delegates (`getData`/`setData`/`setLocalData`/`setMyAwareness`/`requestUpdate`/`unregister`). It holds no state and duplicates no logic; it exists so `register()` can hand back a stable object you can read/write from outside the view (e.g. a form submit) without exposing the internal handler. So there's nothing to merge here — but the surrounding plumbing does have real consolidation opportunities, tracked in #254.

---

packages/playhtml/src/index.ts
spencerc99: Agreed — `register()` already ends at `setupPlayElementForTag`, so collapsing it into one "view path vs imperative path" entry point is the right direction. Per your other note about holding the simplification work until the approach is confirmed, I've deferred this (and the related cleanups) to #254 rather than churn the runtime diff before merge.

---

packages/playhtml/src/index.ts
spencerc99: Both points captured in #254 (items 2 and 3): generalizing the id-based `register(...)` ergonomics to imperative can-play, and reconsidering the implicit `Object.assign(element, init)` stamp in favor of passing `init` to the handler directly. Deferred out of this PR to keep the diff stable for merge.

---

packages/playhtml/src/index.ts
spencerc99: Yes — `define()` already writes to the same `capabilitiesToInitializer` registry that `init({ extraCapabilities })` uses, so the validate → register → upgrade-existing logic can be factored into one shared helper. Tracked as item 4 in #254.

---

packages/playhtml/src/index.ts
spencerc99: Both observe `{ childList, subtree }` and reconcile, so there's likely a shared observer/utility to extract — caveat that their purposes differ (view-descendants binds/tears down capability handlers; can-mirror syncs DOM ↔ shared state), so it needs care. Tracked as item 5 in #254 (flagged highest-risk since it touches existing can-mirror).

---

apps/docs/src/content/docs/custom-elements.mdx
spencerc99: Done in 3f63929. Restructured the whole page into two top-level sections: **Vanilla HTML** (can-mirror → can-play → experimental reactive view API) and **React** (withSharedState only). The intro now leads with the comparison you described — can-mirror syncs everything and the state *is* the DOM; can-play gives fine-grained control over a `data` model. Same example concepts (counter, lamp, guestbook, presence, etc.) are reused in both sections so they mirror each other.

---

spencerc99: The two sections now use the **same example concepts** so they read as the same thing in each framework. Making the live demo *islands* share runtime state across the Vanilla and React sections is more involved (they'd need the same room/element id wired across separate React islands), so I left that out of this pass — want me to file it as a follow-up, or wire it up before merge?

---

apps/docs/src/content/docs/custom-elements.mdx
spencerc99: Fixed by the restructure — the reactive `view` API (the lit-html one) now lives entirely inside the **Vanilla HTML** section, so there's no more lit-html copy shown above a Vanilla/React toggle. The page is split by framework at the top level instead of per-block tabs, so the framework choice is made once, well above any example.

---

apps/docs/src/content/docs/data/data-essentials.md
spencerc99: Done in 42795f9 — the "derive, don't store" example now shows both vanilla forms: the imperative `updateElement` and the reactive `view`, so it's clear you can derive at render time either way.

---

PR #196, opened 2026-06-22
react: avoid shared-state setup loops after room changes

spencerc99:
## Summary

- Prevent `CanPlayElement` from re-running `setupPlayElement` when synced data only updates React state.
- Keep per-render DOM prop refreshes intact, but gate element re-registration on meaningful setup input changes.
- Add a React regression test for synced data updates that should not re-register the element.

## Verification

- `bun run -C packages/react test`
- `bun run -C packages/react build`
- `bun build-site`
- `bun smoke` (rerun with local process/temp escalation after sandbox tempdir blocked Playwright webServer startup)
- Built and browser-tested `spencers-website`: `/` navigates to `/about`, `window.playhtml.roomId` changes from `...-%2F` to `...-%2Fabout`, and no React maximum-depth loop appears.

## Docs

No public docs update: this is an internal React lifecycle fix and does not change the published API.

### DIFF
--- a/.changeset/quiet-paths-navigate.md
+++ b/.changeset/quiet-paths-navigate.md
@@ -1,0 +1,3 @@
+"@playhtml/react": patch
+
+Prevent React shared-state components from re-registering their playhtml element when synced data updates only React state, avoiding render loops after room-change rebinds.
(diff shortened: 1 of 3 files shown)


spencerc99:
Closing this in favor of the smaller main-based core PR: #198.

The React setup-loop cleanup is not part of the SPA init-options fix now that `playhtml.init()` is staying static.

PR #214, opened 2026-06-26
partykit: circuit-break misconfigured bridge pairs

spencerc99:
## Summary

A shared-element **source** room stuck on a stale reset epoch had every `apply-subtrees` rejected by its consumers — but the consumer always responded `ok: true`, so the source never learned and re-fanned-out on **every** doc update. With ~45 subscribers and continuous doc mutations, this produced a retry storm (50+ rejected applies/min per source, 1000+/min site-wide) that kept rooms perpetually hot and unable to self-heal by cold-reloading.

This PR makes bridge rejection **visible** and adds a **per-peer circuit breaker** so a misconfigured pair stops being stormed.

## What changed

- **`ApplySubtreesResponse.applied`** — the receiving room now reports `applied: false` when it rejects an apply (stale reset epoch, or transient persistence-unavailable mode). Optional + backward-compatible: older callers reading only `ok` keep working; an absent `applied` is treated as applied.
- **`bridgeHealth.ts`** — a small in-memory per-peer circuit breaker. After `BRIDGE_CIRCUIT_OPEN_THRESHOLD` (5) consecutive rejected applies to a peer, the source stops sending to *that peer*. Any accepted apply resets the streak.
- **`flushBridgeUpdates`** now routes both fan-out directions (source→consumer and consumer→source) through a single `sendBridgeApply` helper that gates on the breaker, inspects the response, and records the result. The first time a pair trips, it logs once (no per-flush spam).
- **Reset semantics** — the breaker is in-memory, so a Durable Object reload (which produces a fresh, correct epoch) re-enables every bridge automatically. A consumer **resubscribe** (a fresh page load) also reopens its circuit, so a genuine new visitor always gets a clean attempt even if the pair had previously tripped.

## Design notes

- **Per source→consumer pair**, not global: one bad peer never blocks a source's healthy bridges.
- **No backoff/timer layer.** The real failure mode is *persistent* staleness (a peer stuck until it reloads), not transient blips — exponential backoff bought nothing here and added state, so it was cut. The circuit-break + in-memory reset is sufficient and simpler.
- State is deliberately **not persisted**: a cold reload should re-enable bridges, since a reloaded DO has a fresh correct epoch.

## Why this matters (incident)

A misconfigured source (`spencer.place`, stuck on a stale epoch) stormed ~45 consumer rooms continuously. Because the source was blind to the rejections, it never backed off; because the storm kept it hot, it never went idle long enough to reload and pick up a corrected epoch. This breaker stops the storm at the source the moment a pair proves persistently bad.

## Scope

Server-only (`partykit/`). No `packages/` change → no changeset, no published-library impact.

## Testing

- `bun test partykit/__tests__` — 71 pass, 0 fail (6 new in `bridgeHealth.test.ts`: threshold, peer isolation, reset-on-accept, reset-on-resubscribe, never-seen peer).
- `bunx tsc -p partykit` — clean.


spencerc99:
## Addressed review findings (commit c361e31)

**Fixed (correctness):**
- **Ungated re-fanout + new-element push** (the dominant altitude issue) — both raw `consumerRoom.fetch("/apply")` paths now route through `sendBridgeApply`, making it the single outbound chokepoint. A stale-epoch peer can no longer be stormed via the receive-then-refanout path (which fires per consumer-edit).
- **Bidirectional key collision** — the breaker is now keyed by `(peerRoomId, direction)`, so a success in one direction can't clear the other direction's reject streak. New test covers this.
- **`res.ok` not checked** — a non-2xx response (e.g. a JSON error envelope without `applied`) is now scored as a failed apply instead of a success.
- **Double `shouldSend` + missing log on network-failure trips** — `recordResult` now returns the open-edge, so the trip logs exactly once on **both** the response and network-error paths.
- **Dead `reason` field** — dropped from `ApplySubtreesResponse` (never read; the rejection reason is already in the adjacent `warn` log).

**Accepted as known tradeoffs (documented, not fixed):**
- **Consumer→source reset asymmetry** — `reset()` on resubscribe reopens the circuit on the source's side; the consumer→source direction lives in the consumer's DO and is reopened by DO reload (the universal reset) rather than a cross-DO resubscribe signal. Building cross-DO reset signaling is more machinery than the failure mode warrants.
- **Transient failures can trip a healthy link** — 5 consecutive transient fetch failures (no interleaved success) will open the circuit. This is the cost of dropping the backoff/half-open layer (deliberately cut as over-engineering for a *persistent*-staleness failure mode). DO reload / resubscribe recover it. If transient bridge flakiness proves common, a half-open probe is the follow-up.
- **Unbounded health map** — naturally bounded by the room's subscriber/reference count (small ints keyed by room id), cleared on reload. Not worth an eviction pass.

Tests: 74 partykit tests pass (9 in `bridgeHealth.test.ts`, including the bidirectional-collision and open-edge cases). `bunx tsc -p partykit` clean. The `party.ts` fan-out routing is verified by inspection (only one raw `/apply` fetch remains, inside `sendBridgeApply`) since instantiating `PartyServer` needs the Cloudflare runtime.

spencerc99:
## Code review

PR #216, opened 2026-06-28
playhtml: separate config from bootstrap via playhtml.configure()

spencerc99:
## Problem

`playhtml.init(options)` fused two intents into one first-call-wins function: **declare config** (cursors, room, host, …) and **bootstrap** (connect, set up elements). That breaks down whenever there's no single top-level owner of init — Astro islands, multi-page apps, multiple React roots. Multiple independent things each call `init()` to *ensure playhtml is running*, but only one of them owns the config. Because init is first-call-wins, an option-less "ensure running" call can win the race and **silently drop** the owner's config.

This is the root cause of the docs site's intermittent cursors/presence: `HeadOverride` called `init({ cursors })`, while every `@playhtml/react` `standalone` island called `init()` (no options) on mount. Whichever ran first decided whether cursors existed — so they appeared only on some loads.

Symptoms of the old model: ordering mattered invisibly; cursors had to be special-cased as the one option a later `init()` could re-apply; the same function meant two different things depending on the caller.

## Change

Separate the two intents. Keep `init(options)` working exactly as today (no breaking change), and add an authoritative, connection-free way to declare config.

- **`playhtml.configure(options)`** — declares config; does **not** connect. Idempotent and framework-agnostic: call it once, up front, from wherever owns the config (a `<head>` script, `PlayProvider`, an island). Connection happens later via `init()` / component mount, which read the declared config.
- **One uniform rule for every option:** the first declaration wins; a later *genuinely conflicting* option warns and is ignored. Passing the same options again — or none — is silent. The conflict check is **lenient and directional** (only keys the caller actually specifies are compared; functions are excluded since closures never compare equal), so the common "pass identical options at every call site" pattern stays quiet.
- **No breaking change:** `init(options)` is unchanged for single-call-site users. `configure()` is purely additive — the tool you reach for when you can't put one init at the top.

This removes the per-option cursor special-case and keeps the [#198](https://github.com/spencerc99/playhtml/pull/198) guarantee that repeated `init()` never rebuilds providers or triggers a re-render loop.

### Supported ordering

`configure()` before `init()`. An option-less `init()` with no prior config connects immediately with defaults; a `configure()` after connection warns (config can't change once connected). On the docs site `configure()` runs synchronously in `<head>` before any island hydrates, so it always wins.

## Consumers

- **`@playhtml/react`**: `PlayProvider` declares config via `configure()` then ensures running (public contract unchanged). The `standalone` path is ensure-running only (no options) — its `init()` becomes a harmless no-op when playhtml is already started; the element is registered via `setupPlayElement` regardless.
- **Docs site**: `HeadOverride` declares config with `configure({ cursors: { enabled: true, room: "domain" } })`, then `init()`. (This supersedes the interim eager-init fix in #215.)
- **Docs**: `reference/init-options.md` documents `configure()` and the first-declaration-wins rule.

## Tests / verification

- Core: **211 tests pass**; both `playhtml` and `@playhtml/react` type-check clean.
- New `init-configure.test.ts` covers: configure-then-init enables config; connect-immediately-with-defaults + warn on post-connect configure; lenient no-warn on identical/empty options; warn on genuine conflict; and a regression guard that repeated `init()`s connect exactly once (no provider churn / #198 loop).
- Verified in a browser against a built docs site: cursors + presence HUD up on **every** load across 4 pages, two-peer presence works, **zero** false-conflict warnings.

## Relationship to #215

[#215](https://github.com/spencerc99/playhtml/pull/215) is the interim, docs-only fix (eager init, no library change) to unblock the docs immediately. This PR is the proper library-level fix. If #215 merges first, this branch rebases on top and the `HeadOverride.astro` change resolves to the `configure()` version here.

### DIFF
--- a/.changeset/configure-init-split.md
+++ b/.changeset/configure-init-split.md
@@ -1,0 +1,7 @@
+---
+"playhtml": minor
+---
+
+Add `playhtml.configure(options)` to declare init options separately from connecting, and make config first-declaration-wins uniformly across all options.
+
+Previously `init(options)` both declared config and connected, first-call-wins — so on a page with no single top-level init (Astro islands, multi-page apps, multiple React roots), an option-less `init()` from one place could win the race and silently drop the config another place intended (e.g. cursors never turning on). Now you can call `configure({ cursors: { enabled: true }, ... })` once, up front, from wherever owns the config; later `init()` / component mounts just ensure playhtml is running and pick up the declared config regardless of order. `init(options)` still works exactly as before when you only have one call site. A later call that passes genuinely conflicting options warns and is ignored (config is locked to the first declaration); passing the same options again, or none, is quiet.
--- a/apps/docs/src/components/HeadOverride.astro
+++ b/apps/docs/src/components/HeadOverride.astro
@@ -18,22 +18,16 @@ const { head } = Astro.locals.starlightRoute;
   // it initialises before playhtml.init awaits window.load.
   import '../scripts/enhance-code-blocks';
 
-  // Initialise playhtml synchronously, here, before any island hydrates.
+  // Declare playhtml's config synchronously, here, before anything connects.
   //
   // The React islands on these pages render @playhtml/react `standalone`
-  // components, which call `playhtml.init()` (with NO options) when they mount.
-  // init() is first-call-wins, so whichever call runs first decides the config.
-  // This <head> script evaluates while the page is still parsing — before the
-  // <body> islands hydrate — so initialising here (not deferred to window.load)
-  // guarantees OUR config (cursors on, domain-wide room) wins the race. When the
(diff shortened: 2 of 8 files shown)


spencerc99:
## Code review

- **P2 [`packages/playhtml/src/index.ts:1194`](https://github.com/spencerc99/playhtml/blob/8e52df9037a7067f765862df672da6619e457eb8/packages/playhtml/src/index.ts#L1194) adds a public `playhtml.configure()` API without the required starter-template updates.** The repo instructions require both `templates/react-starter/` and `templates/html-starter/` to be updated whenever a public core API changes, including exported functions and `playhtml.init` options. This PR exposes `configure` on the public singleton, but the changed-file list has no files under either starter template. Add the matching starter changes and browser verification so copy-paste starter usage stays current with the new config flow.

PR #225, opened 2026-06-28
Publish init.js drop-in so unpkg.com/playhtml/init.js works

### LINKED ISSUE #222: Add init.js alias so unpkg.com/playhtml/init.js works as a drop-in
## Problem

The docs (and the natural user expectation) reference a drop-in script at `https://unpkg.com/playhtml/init.js`, but that file has **never been published**. The build emits `dist/init.es.js` (a 2-line module that imports `playhtml` and calls `init({})`), and the package `files`/`exports` only expose `dist/`. So the short URL 404s.

PR #220 worked around this by pointing the docs at the verbose `https://unpkg.com/playhtml/dist/init.es.js`. This issue is the proper fix: make the friendly short URL resolve.

## Goal

`https://unpkg.com/playhtml/init.js` should load and run `playhtml.init({})`, so the Path A drop-in in `getting-started.mdx` can use the short, memorable URL.

## Approach

unpkg resolves real files on disk in the published tarball (it does not consult bundler-style `exports` subpath maps for bare paths like `/playhtml/init.js`). So the fix needs an actual `init.js` file in the published package root, plus an `exports` entry for correctness when imported via a bundler.

Concretely:
1. Emit (or add) a root-level `init.js` in the published package that does the same thing as `dist/init.es.js` (`import { playhtml } from "./dist/playhtml.es.js"; playhtml.init({});`). Either change the vite build `fileName` to also output `init.js` at the publish root, or add a tiny committed `init.js` re-export and include it in `package.json` `files`.
2. Add `"./init.js": "./init.js"` (or `./dist/init.es.js`) to the `exports` map.
3. Add `init.js` to `package.json` `files` if it's a committed file.
4. Verify against a real unpkg URL after publish (the build hash path resolves) — `curl -sI https://unpkg.com/playhtml@<next-version>/init.js` should 200.

## Acceptance criteria

- [ ] `https://unpkg.com/playhtml/init.js` returns the init module (200, not 404) after the next release.
- [ ] Changeset added (`playhtml` patch).
- [ ] After release, optionally simplify `apps/docs/src/content/docs/getting-started.mdx` Path A back to `https://unpkg.com/playhtml/init.js` (and the two prose mentions).
- [ ] No regression to the existing `dist/init.es.js` path.

## Context

- Build config: `packages/playhtml/vite.config.ts` (the `fileName` fn emits `init.${format}.js` for the `init` entry).
- `packages/playhtml/package.json` `files` / `exports`.
- Discovered while auditing docs examples in PR #220.

spencerc99:
## Summary

`https://unpkg.com/playhtml/init.js` has never resolved (404) — the build only emits `dist/init.es.js`, and the package `files`/`exports` only exposed `dist/`. unpkg resolves real files in the published tarball for bare paths like `/playhtml/init.js`, so a real file must exist at the package root. This adds it.

Closes #222.

## What I did

- Added a committed `packages/playhtml/init.js` at the package root that imports the built bundle and boots playhtml with defaults:
  ```js
  import { playhtml } from "./dist/playhtml.es.js";
  playhtml.init({});
  ```
  (Includes the required 2-line `// ABOUTME:` header since it's hand-authored.)
- Added `"init.js"` to `package.json` `files` so it ships in the tarball.
- Added `"./init.js": "./init.js"` to the `exports` map so bundler imports of `playhtml/init.js` also resolve.
- Added a changeset (`playhtml` patch).

No change to `dist/init.es.js` — the existing verbose path still works.

## How I verified

- Built the package clean (`tsc && vite build` both pass; `dist/init.es.js` still emitted).
- `npm pack --dry-run` confirms `init.js` is in the published tarball:
  ```
  62B    dist/init.es.js
  16.6kB dist/main.d.ts
  372.4kB dist/playhtml.es.js
  2.1kB  dist/style.css
  238B   init.js
  1.6kB  package.json
  total files: 6
  ```
- Tests: `vitest run --no-typecheck` → **207 passed (24 files)**, no regressions.

## Docs

Per the issue, docs are intentionally unchanged in this PR. After release, `apps/docs/src/content/docs/getting-started.mdx` Path A (and the two prose mentions) can be simplified from `https://unpkg.com/playhtml/dist/init.es.js` back to the short `https://unpkg.com/playhtml/init.js`.


spencerc99:
## Code review

- **P1 [`packages/playhtml/package.json:5`](https://github.com/spencerc99/playhtml/blob/65fbbf4c610ddbe344b026d44fd14f4ec25b44c7/packages/playhtml/package.json#L5) rolls the package version backwards.** The PR head sets `playhtml` to `2.10.1`, while the base side of this diff is `2.11.0`. Because the release flow applies this changeset starting from the checked-in package version, merging this would produce a lower `playhtml` release line than current `main` instead of a patch on top of `2.11.0`. Restore the package version from current `main` before the changeset is applied.

spencerc99:
## Code review

- **P1 [`packages/playhtml/package.json:5`](https://github.com/spencerc99/playhtml/blob/8d85b19a16a69b1b7d0c5a366581c29b1bb61e93/packages/playhtml/package.json#L5) is still based on stale package metadata and would roll back current public package state during conflict resolution.** This head sets `playhtml` to `2.11.3` and its `files`/`exports` lists omit `leafEditor`, while current [`main`](https://github.com/spencerc99/playhtml/blob/8808748b11f5bb40e3a0c546ab38cadce5523264/packages/playhtml/package.json#L5) is `2.13.2` and publishes the `./leafEditor` entry point. Rebase or resolve the conflict by preserving current `main`'s version, files, and exports, then add only `init.js` plus its export.

Issue #223, opened 2026-06-28
Integration-test harness: actually render & assert docs code examples


spencerc99:
## Problem

The docs (`apps/docs/`) show **vanilla HTML/JS** code snippets to readers, but the live demos on those pages are **React** components. The two can drift apart: a demo can keep working while the copy-paste snippet shown to readers is broken. PR #220 found exactly this — three snippets were broken (a `can-duplicate="#id"` that silently failed, a reset selector matching an attribute the library never sets, and a `dist/init.es.js` path), and none were caught because nothing executes the snippets.

We want an automated test that **extracts each vanilla snippet, renders it with the real library, drives the interaction, and asserts the behavior** — a true integration test so a broken example fails CI.

## Chosen approach (decided with maintainer)

**Render + assert behavior**, not just static linting. Example shape:

```ts
test('can-toggle snippet from capabilities.mdx', async () => {
  mountSnippet(snippet);            // inject the fenced ```html block into the DOM
  await playhtml.init();
  btn.click();
  expect(btn.classList.contains('toggled')).toBe(true);
});
```

## Feasibility (already scouted)

This is very achievable because the infra already exists:

- **`packages/playhtml/src/__tests__/vitest.setup.ts`** already fakes the `y-partyserver` provider (`vi.mock("y-partyserver/provider", ...)`) and `y-indexeddb`, so **`playhtml.init()` runs fully offline in jsdom** with no live PartyKit server. Existing tests (`main.basic.test.ts`, `duplicate-element-ref.test.ts`) already call `playhtml.init({})` and drive clicks in jsdom.
- `vitest.config.ts` aliases `@playhtml/common` to source.
- So the missing pieces are just: (1) a snippet extractor, (2) a mounter, (3) per-capability assertions.

## Suggested implementation

1. **Snippet extractor**: parse the `.mdx`/`.md` files in `apps/docs/src/content/docs/` and pull fenced code blocks. Tag each block with its language (`html`, `tsx`, `js`) and a nearby anchor (heading / capability name) so failures point back to the source. Only `html`/`js` vanilla blocks are executed; `tsx`/React blocks are out of scope for v1 (the live React demos already exercise those paths).
2. **Mounter**: a helper that takes an HTML snippet string, injects it into `document.body`, strips/handles any `<script>` per the snippet's intent, runs `playhtml.init()`, and returns handles to the elements.
[...]

PR #228, opened 2026-06-28
Add docs-example integration tests (render & assert vanilla snippets)

spencerc99:
## Summary

Adds a docs-example integration-test harness in `packages/playhtml` that extracts vanilla HTML/JS snippets from `apps/docs`, renders them in jsdom with the **real** `playhtml` library, drives the interaction, and asserts behavior — so a broken copy-paste example fails CI instead of silently shipping.

Addresses #223 (prototype scope: extractor + mounter + two capabilities).

## Files added

Both start with `// ABOUTME:` headers, no emoji.

- `src/__tests__/docs-snippets.ts` — the extractor (`extractSnippets` / `vanillaSnippets`), `mountHtml` mounter with handler cleanup, `withConsoleCapture`, and the `isExecutable` opt-out predicate.
- `src/__tests__/docs-examples.test.ts` — 4 tests: extractor unit tests, the opt-out convention, and behavior assertions for `can-toggle` and `can-duplicate` pulled from the **real** `capabilities.mdx` snippets.

## Key decisions

- Reuses the existing provider fake in `packages/playhtml/vitest.setup.ts` — no new mocking. `playhtml.init()` runs fully offline in jsdom.
- Co-located in `packages/playhtml` so docs files load by relative path and the existing vitest config applies.
- Opt-out convention: a ` ```html ignore-test ` fence flag OR a `/* ... */` placeholder in the body marks a snippet as non-executable. Only `html`/`js` blocks run; `tsx` is out of scope for v1 (the live React demos already exercise those paths).
- Inline `<script>` tags in snippets are left inert (jsdom doesn't run innerHTML scripts; doc scripts bind to full-page ids) — documented in `mountHtml`.

## Coverage

- `can-toggle` — clicks the real docs snippet, asserts the `toggled` class flips on/off.
- `can-duplicate` — clicks the real docs snippet, asserts a clone appears (`[id^='bunny-template-']`).

Both assert pristine console output (no errors/warnings), per the project's clean-output rule.

## Proof it catches breaks

Temporarily repointed the `can-toggle` assertion at a fake class → the suite went red (1 failed / others passed); reverted → all green. So a genuinely broken snippet fails here.

## Stacking note

This was rebased onto `main` after #220 merged, so the `can-duplicate` test exercises the now-correct bare-id snippet (`can-duplicate="bunny-template"`). The harness passing on the post-#220 docs is the end-to-end confirmation that #220's `#`/selector fix works through real docs markup.

## Test results

New file: 4 passed. Full suite: **26 files / 215 tests, all passing**, no regressions. No changeset (test-only, no shipped-behavior change). The only non-test console line during the run is an unrelated pre-existing Sass `legacy-js-api` deprecation warning from the build pipeline.

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -32,6 +32,9 @@ jobs:
     - name: Run core tests
       run: cd packages/playhtml && bun run test
 
+    - name: Run docs tests
+      run: cd apps/docs && bun run test
+
     - name: Run react tests
       # 2 pre-existing pathname-prop failures (mock-restore quirk) are
       # tracked separately. Reports failures but does not block merge.
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -7,6 +7,7 @@
     "dev": "astro dev --port 4321",
     "start": "astro dev",
     "build": "astro build",
+    "test": "vitest run --no-typecheck",
     "preview": "astro preview",
     "astro": "astro"
   },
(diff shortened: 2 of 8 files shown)


spencerc99:
## Code review

spencerc99:
why are the tests inside packages/playhtml? These should not be inside the core package

PR #232, opened 2026-06-28
playhtml: avoid duplicate can-grow hover listeners

spencerc99:
## Summary
- register can-grow mouseenter and mouseleave handlers once per mounted element
- guard document key listener registration so repeated enter events do not stack handlers
- add a regression test for repeated can-grow hover events

## Verification
- red: `bun run test -- src/__tests__/main.basic.test.ts -t "does not add a mouseleave listener"` failed with two mouseleave listeners
- green: `bun run test -- src/__tests__/main.basic.test.ts -t "does not add a mouseleave listener"`
- `bun run test -- src/__tests__/main.basic.test.ts`
- `bun run build` in `packages/playhtml` after adding a local ignored node_modules symlink for worktree resolution; output includes the existing API Extractor TypeScript-version warning
- `git diff --check`

## Docs and starters
No docs or starter updates needed: this does not change can-grow API, attributes, classes, examples, or setup.


spencerc99:
## Code review

PR #236, opened 2026-06-29
Fix can-mirror losing children with identical or mixed children

spencerc99:
## Problem

`can-mirror` is supposed to mirror an element's full DOM subtree across clients, but it dropped or corrupted children whenever the subtree contained **multiple identical children** (e.g. a grid of identical `` tiles) or a **mix of text and element nodes**.

The reported symptom: a container with several identical cat images — removing the *last* child removed *all* of them on connected clients.

### Root cause

Two separate defects, both stemming from matching DOM nodes to stored state by value rather than by position:

1. **Observe side (`updateChildList`).** Children were reconciled by value-equality (`areStatesEqual`). The add path refused to store a child that "looked the same" as an existing one, so three identical children deduped to a single state entry. With state holding 1 entry but the DOM holding 3, removing one child emptied the lone entry, and the empty state then wiped every child via `updateChildrenFromState`.

2. **Apply side (`updateChildrenFromState`).** Remote state was applied to the DOM strictly by position, assuming the DOM node and the state entry at each index were the same *kind* of node. When a sibling removal shifted things so a text node lined up against element state (or vice versa), it called `updateAttributesFromState` on a text node and threw (`element.attributes` is undefined), blanking the update.

## Fix

1. `updateChildList` now resyncs the entire `children` array from the live DOM by position on each childList mutation (the DOM is the source of truth for order and count). Uses `splice(0, len, ...new)` — the Yjs-proxy-safe in-place array replacement already used by `syncInputState` in this file. The observer runs with `subtree: false` and a `target === element` filter, so this only ever sees direct-child changes.
2. `updateChildrenFromState` now replaces a positional child when its kind (nodeType, or element tag) no longer matches the target state, instead of trying to update it in place. Node-creation logic is factored into a shared `createNodeForState` helper.

## Testing

New `packages/playhtml/src/__tests__/can-mirror.test.ts` drives the real `onMount` MutationObserver → shared state → `updateElement` → DOM path in jsdom (real MutationObserver, real DOM — no mocked behavior), covering:
- removing the last of several identical children removes only one
- same-client round-trip after removal
- sequential removals stay consistent
- add-then-remove (the cleanest reproduction; failed before the observe-side fix)
- remote-client sync after removing an identical child
- mixed text + element children preserve order/count after a middle element is removed (this one surfaced the apply-side crash)

All 6 pass. Type-check clean.

**Live-verified** in a browser against `website/test/mirror.html` with full playhtml init + sync round-trip: adding 3 identical children then removing the last leaves 2 (not 0); removing the first element child from a mixed text/element list leaves the correct remaining nodes in order with no console errors.

## Docs

No docs change needed — this is an internal correctness fix to an existing capability with no change to the public attribute/API surface.

## Changeset

Included (`@playhtml/common` patch).

### DIFF
--- a/.changeset/can-mirror-identical-children.md
+++ b/.changeset/can-mirror-identical-children.md
@@ -7,3 +7,5 @@ Fix `can-mirror` dropping or corrupting children when an element has multiple id
 The child-list sync matched DOM nodes to stored state by value-equality, so repeated children (e.g. several identical images or list items) were deduplicated into a single entry. Removing one child then emptied the lone entry and wiped every child on connected clients. Child-list mutations now resync from the live DOM by position, preserving order and count.
 
 Applying remote state also reconciled children strictly by position, which threw when a position's node kind changed (e.g. a text node lining up against element state after a sibling was removed). Mismatched nodes are now replaced rather than updated in place.
+
+Applying remote state to an element that was emptied remotely now correctly removes its stale children, instead of leaving them in place.
--- a/.changeset/can-mirror-identical-children.md
+++ b/.changeset/can-mirror-identical-children.md
@@ -1,0 +1,9 @@
+---
+"@playhtml/common": patch
+---
+
+Fix `can-mirror` dropping or corrupting children when an element has multiple identical children or a mix of text and element children.
+
+The child-list sync matched DOM nodes to stored state by value-equality, so repeated children (e.g. several identical images or list items) were deduplicated into a single entry. Removing one child then emptied the lone entry and wiped every child on connected clients. Child-list mutations now resync from the live DOM by position, preserving order and count.
+
+Applying remote state also reconciled children strictly by position, which threw when a position's node kind changed (e.g. a text node lining up against element state after a sibling was removed). Mismatched nodes are now replaced rather than updated in place.
--- a/packages/common/src/canMirror.ts
+++ b/packages/common/src/canMirror.ts
@@ -461,9 +461,10 @@ function updateElementFromState(
     updateAttributesFromState(element as HTMLElement, newState);
     applyFormState(element as HTMLElement, newState.formState);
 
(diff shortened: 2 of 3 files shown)


spencerc99:
Expanded the tests into a thorough behavioral suite (31 tests), organized by capability surface and modeled on the can-mirror playground:

- **attributes** — add / change / remove, native `open` on `<details>`, and confirmation that ephemeral `data-playhtml-hover`/`-focus` are filtered out of persisted state
- **child add/remove** — add, remove, identical children (no dedup), clear-all, reorder
- **mixed text + element children** — order preserved on middle-element removal; node-kind and tag replacement on apply
- **character data** — syncs via `input` event; documents the `subtree: false` boundary (a bare text-child edit with no input event is not observed)
- **form state** — checkbox, radio, text input, textarea, select `selectedIndex`
- **contenteditable** — text edit and list-item add via `input`
- **awareness** — hover/focus attributes reflect any peer
- **apply-side robustness** — no-op when already equal, stale attribute removal, deep nested subtree
- **feedback-loop safety** — applying one's own echoed state doesn't re-trigger the observer
- **sequential convergence** — a multi-step edit walk, including a nested-text change carried along by a root-level re-snapshot

The tests use a two-client harness (source observes → state crosses the wire → sink applies) over the real observe/apply path with a real MutationObserver and DOM. They don't model Yjs CRDT merge semantics — that's runtime-level, out of scope for canMirror's pure DOM↔state logic (noted in the file header).

**The suite found one more bug, now fixed:** applying remote state to an element that was emptied remotely left its stale DOM children in place, because `updateElementFromState` skipped child reconciliation when the target state had zero children. It now always reconciles. Verified live: populating then clearing a `can-mirror` element syncs to empty with no console errors.

spencerc99:
## Code review

PR #246, opened 2026-06-30
Scope can-play composition with built-ins

spencerc99:
## Summary
- Scope bare `can-play` DOM props away from built-in capabilities when an element uses both.
- Add regression coverage for `can-play` + `can-move` composition.
- Document the supported composition behavior and add a patch changeset.
- Fix TypeScript nullability around SyncedStore/page-data proxy access found during package build.
- Leave starter templates unchanged because their current examples do not demonstrate multi-tag composition and remain copy-pasteable.

## Root cause
The runtime read generic DOM element props for every capability tag on a node, so `can-play` props like `defaultData` and `updateElement` could override a built-in initializer such as `can-move`.

## Verification
- `./node_modules/.bin/vitest run --no-typecheck` in `packages/playhtml` (31 files, 284 tests passed; existing perf stdout and expected `can-duplicate` stderr)
- `bun run build` in `packages/playhtml` (passed with existing Sass legacy API warning)
- `/Users/spencerchang/Projects/playhtml/node_modules/.bin/astro build` in `apps/docs` (passed with existing Vite/Astro warnings)
- `git diff --check`

### DIFF
--- a/.changeset/composed-candle.md
+++ b/.changeset/composed-candle.md
@@ -1,0 +1,5 @@
+---
+"playhtml": patch
+---
+
+Keep `can-play` element properties scoped to `can-play` when an element also uses a built-in capability, so custom state can compose with built-in behavior such as `can-move`.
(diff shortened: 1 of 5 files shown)


spencerc99:
## Code review

PR #248, opened 2026-06-30
Fix can-mirror registration for mirrored dynamic elements

spencerc99:
## Summary

- Register PlayHTML capability descendants after `can-mirror` applies shared DOM state, so dynamically mirrored children become interactive without a delayed `setupPlayElement` re-walk.
- Add a regression test shaped like the musical chairs example: a `can-mirror` container receives a `.chair` child with `can-spin`, `can-toggle`, and a `/red-stool.png` image, then verifies the mirrored chair gets live handlers.
- Add a patch changeset and a short docs note for nested capability children inside `can-mirror`.
- Capture expected `can-duplicate` missing-template logs in the existing test so package test output stays clean.

## Root Cause

`can-mirror` correctly applied the child DOM from shared state, but runtime element setup only happened during the top-level discovery pass or an explicit `playhtml.setupPlayElement(...)` call. Descendants created by `can-mirror` during load were inserted as plain DOM nodes, so their own `can-spin`, `can-toggle`, or other capability attributes were not registered.

## Validation

- `bun run test -- src/__tests__/main.basic.test.ts`
- `bun run test -- src/__tests__/can-mirror.test.ts`
- `bun run build` in `apps/docs` (passes with existing Vite/Astro warnings)
- `bun run lint` still fails in unrelated extension/worker TypeScript surfaces after building `@playhtml/extension-types`: trail settings props, browser storage/data transfer types, worker runtime globals, page metadata `cf` fetch typing, and WXT manifest typing.

### DIFF
--- a/.changeset/mirrored-chair-setup.md
+++ b/.changeset/mirrored-chair-setup.md
@@ -1,0 +1,5 @@
+---
+"playhtml": patch
+---
+
+Set up playhtml capability descendants created by can-mirror state so mirrored dynamic elements become interactive without an extra DOM setup pass.
--- a/apps/docs/src/content/docs/capabilities.mdx
+++ b/apps/docs/src/content/docs/capabilities.mdx
@@ -458,6 +458,8 @@ Click "+" to append a new `<li>`. The child addition itself mirrors, so new read
 ">add entry</button>
 ```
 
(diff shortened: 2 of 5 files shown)


spencerc99:
## Code review

PR #260, opened 2026-07-02
Fix Wikipedia link patina for absolute article links

spencerc99:
## Summary

- Normalize Wikipedia article links before looking up link patina data, so same-origin relative, protocol-relative, and absolute article URLs all map to the stored `/wiki/...` path.
- Keep namespace and offsite links excluded from link glow rendering.
- Add a regression test for the Wikipedia Main Page-style protocol-relative home link.
- Add the extension pending-release note and record the extension-facing `extension/PENDING.md` requirement in the repo agent instructions.

## Root cause

Wikipedia is now serving some article pages with article anchors like `//en.wikipedia.org/wiki/Wikipedia` instead of `/wiki/Wikipedia`. `LinkGlowManager` only scanned `a[href^="/wiki/"]`, so those same-origin article links never got registered or rendered with patina even though the stored click data still existed.

## Validation

- `bun run test -- run` from `extension/`: 43 files passed, 286 tests passed.
- `bun build-extension`: WXT Chrome MV3 production build passed.

Note: the first attempted verification command used the documented `bun run -C extension test -- run` shape, but this Bun version rejected `-C`; reran the same test script from `extension/` successfully.


spencerc99:
## Code review

PR #263, opened 2026-07-03
Fix dev inspect cleanup and clarify can-mirror scope

spencerc99:
## Summary

- Clean up development inspector highlight classes when development mode is torn down and keep those internal classes out of `can-mirror` shared state.
- Keep `can-mirror` scoped to the element that declares it: attributes on that element, its direct child list, and form/contenteditable input state.
- Document the non-recursive `can-mirror` boundary and update the manual mirror test page to show nested elements opting in with their own `can-mirror`.

## Root Cause

Development inspector classes were regular DOM classes, so mirrored elements could persist them like user-authored classes. While investigating the removal path, the tempting recursive observer fix created broader ownership problems: parent mirrors could capture nested capability output, duplicate ownership with child mirrors, and overwrite unrelated descendant edits. The final behavior keeps ownership explicit.

## Validation

- `bunx vitest run --no-typecheck src/__tests__/can-mirror.test.ts -t "nested attribute removal|nested child removal|direct text-child mutation|child resyncs" --reporter verbose`
- `bun run test -- src/__tests__/can-mirror.test.ts src/__tests__/main.basic.test.ts src/__tests__/duplicate-ids.test.ts`
- `bun run build-site`
- Browser smoke on `/test/mirror.html` for root removal controls, explicit nested `can-mirror`, and hover console logs


spencerc99:
## Code review

- **P2 [`apps/docs/src/content/docs/capabilities.mdx:382`](https://github.com/spencerc99/playhtml/blob/9d3f79438175a7cfe69d39fcdb97bc5082aa9c91/apps/docs/src/content/docs/capabilities.mdx#L382) changes the public `can-mirror` contract without handling the required starter-template audit.** The repo instructions require both starter templates to be updated for public core API/capability behavior changes, or for the PR summary to explain why no starter change is needed. This PR documents a narrower `can-mirror` boundary here, and the starters do expose `can-mirror` examples (`templates/html-starter/index.html:62`, `templates/html-starter/index.html:176`, `templates/react-starter/src/App.tsx:119`), but neither starter is in the PR file list and the PR summary has no starter-template rationale. Update the affected starter examples, or add the explicit "no starter change needed" rationale to the PR summary if the existing examples are still correct.

PR #264, opened 2026-07-03
Identity, authentication & permissions system

spencerc99:
## Identity, authentication & permissions for playhtml

Adds a full identity + auth + permissions system to the library, integrated with the "we were online" extension.

### What it does

- **Stable identity for everyone.** Anonymous users get a real ECDSA P-256 keypair (non-extractable private key in IndexedDB) whose public key is their `pk_…` pid — the same format the extension uses. Extension identities take precedence via the existing `playhtml:configure-identity` bridge.
- **Declarative permissions.** A `permissions` init option (and `/.well-known/playhtml.json`) declares roles and rules gating element writes. Checks are synchronous via `playhtml.can()` / the `useCan()` React hook; `usePlayerIdentity()` reports `verified` and `roles`; denied writes fire a `permissiondenied` event.
- **Real server enforcement.** When a domain publishes `/.well-known/playhtml.json`, the partykit server enforces the rules: connections prove key ownership through a challenge–response handshake over the existing WebSocket (one signature per connection, session-token resume, zero per-message crypto), and writes to gated elements are server-mediated — including keyed-collection entry rules (`create`/`update`/`delete` with a built-in `creator` role and server-stamped `createdBy`). Sites that don't opt in pay no cost.

### Earned roles (experimental)

Roles can also be **earned by showing up**: `{ "days": N }` grants a role to any verified identity the server has seen on N distinct days in the room (counted server-side, once per day — uninflatable by reconnecting or clearing storage). `days` is one of the built-in server-attested **counters**; `{ "sessions": N }` works the same way. Totals are reported in `playhtml.me.counters`.

This started as a hardcoded `{ visits: N }` and was generalized into a counter primitive so the earned-standing mechanic doesn't bake one metric into the wire format or persisted server storage — new counters can slot in later with no migration. **Flagged experimental in the docs**; the counter vocabulary may still change. Explicit key-list roles are stable.

### Examples

- **The village guestbook** (`website/guestbook.html`) — a returning/regular/keeper ladder demonstrating earned standing as client-UX flavor over server-enforced owner-key + moderator powers.
- Plus permission-gated example pages and an auth E2E smoke test (`smoke:partykit:auth`).

### Testing

- playhtml **346/346**, partykit auth policy suite green, react auth tests green.
- Node-based E2E smoke test exercises the full protocol against a live wrangler dev server: handshake, write-gating, entry ownership, backstop revert, session resume, and the earned-role ladder (time-compressed via `AUTH_DAY_MS`).

### Notes

- Existing anonymous users' pids change once when the legacy random-hex publicKey is upgraded to a real key (accepted in the design).
- Docs: new **Permissions** page under Advanced.


spencerc99:
## Code review

- **P1 [`partykit/party.ts:2463`](https://github.com/spencerc99/playhtml/blob/fb8941e7eccb9514332c169305ce2d4cd9dbfe42/partykit/party.ts#L2463) skips arming the raw-Yjs enforcement backstop when the well-known config comes from Durable Object storage.** `getPermissionsConfig()` returns the fresh stored config directly at lines 2463-2465, but `armGatedEnforcement(config)` is only called on the fresh network-fetch path at lines 2473-2475. On a cold Durable Object instance with a still-fresh cached `/.well-known/playhtml.json`, the normal `gated_write` custom message path still checks permissions, but the `play.observeDeep` backstop that reverts direct Yjs writes is never attached. A client that bypasses the library and sends raw Yjs updates can therefore mutate gated elements until the cache expires and the fetch path happens to arm enforcement. Make `getPermissionsConfig()` arm enforcement before every non-null cached return, or centralize non-null config returns through an idempotent helper.

spencerc99:
I checked the signing path here because of the private/public identity boundary issue from #286. The high-level shape is right: page code asks the content script for a signature, the content script asks background, background validates the `playhtml-auth-v1` payload and binds `payload.origin` to `sender.url`, and the server verifies the same signed payload against the room domain. That lets the page get a session-specific signature without receiving the private key.

The part I think we should change before merging is the extension key storage/API boundary:

- `extension/src/entrypoints/background.ts` generates the extension keypair as extractable, exports the private key as JWK, and stores it at `playerIdentity.privateKey`.
- `GET_PLAYER_IDENTITY` returns the stored object wholesale.
- `extension/src/entrypoints/content.ts` then dispatches that identity through `playhtml:configure-identity`, which can carry the raw JWK across the DOM bridge before the library sanitizes it.

That means the signing bridge is mostly doing the right thing, but the identity retrieval path still makes the raw private key grab-able/passable as normal data. I think the fix should be structural rather than relying on every caller to remember to sanitize:

1. Make the only shareable identity object `playerIdentity.public` / `PlayerIdentity`.
2. Remove private key material from any runtime message, content-script state, popup/profile API, DOM event detail, presence payload, or synced/public identity object.
3. Store extension private signing material separately, ideally as a non-extractable `CryptoKey` in extension IndexedDB rather than exported JWK in `browser.storage.local`.
4. Keep background as the only signer via `SIGN_PLAYHTML_CHALLENGE`. That message should return only `{ signature }` or `{ error }`, after validating payload shape and origin.

So I would not remove browser signing. The extension still needs to sign secure messages. But we should remove the ability to fetch or pass around `privateKey` as data; code should only be able to ask the background signer to sign a narrowly validated PlayHTML auth payload.

spencerc99:
## Code review

- **P1 [`partykit/auth.ts:494`](https://github.com/spencerc99/playhtml/blob/f504ddf68f511cdc879dc657ed06bde8daaa8d79/partykit/auth.ts#L494) misses brand-new gated data when a raw Yjs update changes a whole tag.** `collectChangedElementIds()` returns `null` for a top-level tag-map change, but the null branch checks only `Object.keys(snapshots)`. A malicious client can therefore add a new tag containing a rule-matched element that has no authoritative snapshot yet; its ID is absent from the candidate list, so the backstop neither removes nor restores it and the unauthorized value persists. When the changed set is unknown, enumerate current IDs under the changed/live tag maps as well as stored snapshots before applying the rule filter.

- **P1 [`partykit/party.ts:2775`](https://github.com/spencerc99/playhtml/blob/f504ddf68f511cdc879dc657ed06bde8daaa8d79/partykit/party.ts#L2775) stores only one authoritative tag snapshot per element ID.** `snapshots[elementId]` is overwritten with a single `{ tag, data }`, while the document legitimately stores the same ID under multiple capability tags (and now under the `__page__` tag). After gated writes to two such tags, a raw write to the first tag causes the backstop to restore only the last tag snapshot, leaving the unauthorized first-tag value in place. Key snapshots and changed candidates by `(tag, elementId)`, or store and restore every tag value for an element ID.

spencerc99:
## Code review

- **P1 — Whole-tag additions still bypass the raw-write backstop.** [`collectChangedElementIds()` still returns `null` when the transaction changes a key directly on the top-level `play` map](https://github.com/spencerc99/playhtml/blob/ec70d5de1f38030314a93b09d09a12ce8a5c3d83/partykit/auth.ts#L437-L465), and [`planGatedReverts()` still turns that unknown change set into only `Object.keys(snapshots)`](https://github.com/spencerc99/playhtml/blob/ec70d5de1f38030314a93b09d09a12ce8a5c3d83/partykit/auth.ts#L488-L507). The no-snapshot removal fix works only when an element ID reaches the candidate set. A newly inserted tag containing a gated ID has neither a snapshot nor a candidate entry, so the backstop schedules no removal. When the change set is unknown, include current element IDs from the changed/live tag maps as well as stored snapshots before resolving rules.

- **P1 — Authoritative snapshots still collapse all tags for an element ID into one value.** [A successful gated write still assigns `snapshots[elementId] = { tag, data }`](https://github.com/spencerc99/playhtml/blob/ec70d5de1f38030314a93b09d09a12ce8a5c3d83/partykit/party.ts#L3239-L3265), overwriting any authoritative value for the same ID under another capability tag or `__page__`. The backstop then [checks and restores only the tag stored in that one snapshot](https://github.com/spencerc99/playhtml/blob/ec70d5de1f38030314a93b09d09a12ce8a5c3d83/partykit/party.ts#L3325-L3337). If the later tag remains authoritative while an earlier tag is changed directly, the equality check returns early on the later tag and leaves the earlier unauthorized value intact. Track candidates and snapshots by tag plus element ID, then restore the exact changed tag value.

PR #266, opened 2026-07-03
Firm up can-move bounds

spencerc99:
## Summary

- Make `can-move-bounds` keep the full element inside its bounds by default.
- Clamp relative to the element's starting position inside the bounds container, so elements that do not start at the container origin behave consistently.
- Normalize persisted out-of-bounds `can-move` positions on mount, and keep explicit `min-visible` settings available for partial overhang.
- Update docs, React prop comments, regression coverage, and changeset release notes.

## Root Cause

The previous bounds math intentionally allowed partial overhang by default, and it assumed the draggable started at the bounds container origin. That made edge behavior inconsistent for elements with a non-zero starting layout position. After tightening the drag clamp, persisted positions from old rooms could still render outside the bounds because initial render applied stored `{ x, y }` directly before any drag happened.

## Docs And Starters

Docs were updated in the capabilities page and React API reference. Starter templates are unchanged because they do not demonstrate `can-move-bounds`, so their current examples remain copy-pasteable.

## Validation

- `bun run test src/__tests__/can-move-bounds.test.ts` in `packages/playhtml`
- `bun run build` in `packages/react`
- `bun run build` in `apps/docs`
- `git diff --check HEAD`

`packages/playhtml` tests still print existing Lit dev-mode warnings and perf output.

### DIFF
--- a/.changeset/firm-move-bounds.md
+++ b/.changeset/firm-move-bounds.md
@@ -4,4 +4,4 @@
 "@playhtml/react": patch
 ---
 
-Make `can-move-bounds` clamp the full element inside its bounds by default, account for the element's starting position within the bounds container, and normalize persisted out-of-bounds positions on mount. Explicit `min-visible` settings can still allow partial overhang when that behavior is wanted.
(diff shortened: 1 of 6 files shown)


spencerc99:
## Code review

PR #278, opened 2026-07-06
Move element awareness and page presence onto the presence transport (closes #240)

### LINKED ISSUE #240: Migrate PlayHTML presence APIs to generic presence transport
## Context

PR #206 moved high-frequency cursor traffic onto the generic PartySocket presence transport. The next migration should move the rest of PlayHTML presence and awareness onto that transport only after cursor stability has production soak time.

## Scope

- Keep Yjs for persistent shared document sync.
- Move `playhtml.presence.setMyPresence`, `onPresenceChange`, and `createPresenceRoom` to the generic presence transport.
- Move element `setMyAwareness` to element-scoped presence channels after the public presence path is stable.
- Decide whether a one-release compatibility bridge is needed for package/server skew.
- Retire Yjs awareness from public PlayHTML APIs once the generic transport has enough soak history.

## Acceptance criteria

- Public presence APIs use the generic presence transport and have tests for reconnect, replay, destroy, and multi-tab identity behavior.
- Element awareness uses generic presence channels and has tests for can-* awareness callbacks.
- User-facing docs are updated where APIs or behavior change.
- Staging validation covers cursor traffic plus custom presence/element awareness before production rollout.

Related: PR #206.

### LINKED ISSUE #273: Move element awareness onto page-scoped presence transport
## Context

PR #272 fixes element awareness syncing by keeping element awareness scoped to the page room instead of following `cursorClient?.getProvider()`. Longer-term, we still want element awareness to use the generic realtime presence transport instead of Yjs awareness, but we need to preserve that page-room scope.

The important distinction is transport vs. scope:

- Cursor presence can use `cursors.room` (`page`, `domain`, `section`, or custom).
- Element awareness must remain scoped to the element's page room, because it is keyed by tag + element id and rendered by page elements.
- The presence transport should be treated as a generic ephemeral-state transport, not as cursor-specific state.

## Proposed design

Use `RealtimePresenceTransport` / `PresenceServer` for element awareness, but connect it to the normalized page room.

- If the cursor presence room equals the page room, reuse the same presence socket.
- If the cursor presence room differs from the page room, open a separate page-scoped presence socket for element awareness.
- Cursor-room changes should not affect element-awareness delivery unless the shared socket needs to split or rejoin.
- Page-room navigation should rebuild the element-awareness transport.

Publish element awareness as page-scoped presence channels. Prefer one channel per tag rather than one channel per element so we do not hit the server's per-connection channel cap:

```ts
channel: `element:${tag}`
value: {
  [elementId]: awarenessValue
}
```

The client can reconstruct the existing runtime shape from page-room presence peers:

- `awareness: V[]`
- `awarenessByStableId: Map<string, V>`

Use the joined player identity/public key for stable ids where available, falling back only when identity is missing.

## Non-goals

- Do not publish page element awareness into a broader cursor room with the page encoded in the channel name. That would leak page-specific element ids across the cursor scope and waste bandwidth.
- Do not make element awareness follow `cursors.room` semantics.
- Do not reintroduce server-side Yjs-awareness pruning for this migration.

## Implementation notes

- Keep the current Yjs-awareness path as a fallback until the transport path is proven in browser tests.
- `setMyAwareness()` should update a local per-tag map and publish `presence.update("element:${tag}", tagMap)`.
- Element removal should delete that element id from the local tag map and republish, or clear the tag channel when empty.
- Navigation should rebuild the page-scoped element-awareness transport only when the page room changes.
- The existing `PresenceServer` already treats `element:` channels as interactive cadence, which is a good fit for hover/focus-style awareness.

## Acceptance criteria

- Two clients on the same page sync element awareness through the presence transport when `cursors.room` is also `page`.
- Two clients on the same page sync element awareness when `cursors.room` is `domain` or another non-page scope.
- Two clients in the same cursor/domain room but on different pages do not see each other's element awareness.
- Closing one browser removes its element awareness from the remaining browser.
- Page navigation rebuilds element-awareness transport correctly.
- Cursor-room-only changes do not orphan or rebuild the page element-awareness listener unnecessarily.
- Browser smoke coverage verifies exact two-client convergence and cleanup after one client leaves.

## Related

- #240
- #272

spencerc99:
Closes #273

## Summary

Element awareness (`setMyAwareness` / `updateElementAwareness`) now syncs over the generic realtime presence transport (`RealtimePresenceTransport` / `PresenceServer`) instead of Yjs awareness, always scoped to the element's **page room** — including when `cursors.room` is `domain` or another non-page scope.

- **New `ElementAwarenessClient`** (`packages/playhtml/src/element-awareness.ts`): publishes the local per-tag map as one `element:<tag>` presence channel per tag (`{ [elementId]: awarenessValue }`), and rebuilds `awareness` / `awarenessByStableId` from page-room presence peers. Stable ids come from the joined identity's `publicKey`, falling back to the connection id.
- **Refcounted per-room transport registry** in `index.ts`: when the cursor room equals the page room, cursors and element awareness share one presence socket; otherwise element awareness opens its own page-scoped socket. Cursor-room-only changes never rebuild or orphan the element-awareness listener; page navigation rebuilds it (transport ownership moved out of `CursorClientAwareness` — it's now a borrower).
- **Yjs fallback preserved**: when WebSocket is unavailable the entire old path (awareness listener, `ensureAwarenessIdentity`, fingerprint short-circuit) still runs, pinned by dedicated tests.
- **No server changes** — `PresenceServer` already treats `element:*` channels as interactive cadence. **No public API changes.**
- Element removal (`removePlayElement`) deletes the id from the tag map and republishes (or clears the channel when empty); extension identity injection re-joins the element transport.

## Validation

- `packages/playhtml`: 371/371 tests (includes new unit tests for the client, transport-sharing tests, and navigation/isolation tests; several were mutation-tested to prove they fail against the bugs they guard)
- `packages/react`: 43/43 tests; `tsc --noEmit` clean; `bun run lint` clean for packages/ + website/ (remaining errors are pre-existing in extension workspaces)
- **Staging browser smoke** (real two/three-browser clients against the deployed staging worker): existing homepage two-client element-awareness sync + cleanup-on-close test passes on the new transport path, and a new test verifies page-scoping with `cursors.room: "domain"` (clients on the same domain cursor room but different pages do not see each other's element awareness)

## Notes for review

- **Mixed-version rollout**: old clients (Yjs element awareness) and new clients on the same page are mutually invisible for element awareness until both are on this version. Ephemeral presence only (e.g. "N people here" counters); no persisted data involved, self-heals as bundles refresh. Called out in the changeset.
- Two review-loop fixes worth knowing about: both joiners on a shared socket now publish the identical presence `page` value (via `getPresencePage`), and local `setMyAwareness` fires `updateElementAwareness` exactly once (the transport branch of `triggerAwarenessUpdate` bare-returns since `setLocalAwareness` already emitted synchronously).
- A handler mounted after a peer's awareness already arrived is seeded with that peer state on the next presence message — exact parity with the pre-migration Yjs fingerprint behavior.
- The `bun.lock` hunk is a version-field sync with the release commit's package.json versions (byproduct of `bun install`), not a version bump.
- Docs audit: no `apps/docs` changes needed — no doc makes claims about the awareness transport, and the public API/attributes are unchanged. Starter templates unaffected for the same reason.
- `website/index.tsx` gains a localhost-only `playhtmlCursorRoom` preview param (validated against the `CursorRoom` literals) so the smoke test can exercise the domain-cursor-room case.


spencerc99:
## Code review

- **P1 [`packages/playhtml/src/element-awareness.ts:57`](https://github.com/spencerc99/playhtml/blob/9ef5393ca2311f525622e3430d5cf7db3d0f709b/packages/playhtml/src/element-awareness.ts#L57) batches every local element's awareness for a tag into one presence value, which crosses the transport value-size cap on realistic pages and stops publishing later awareness.** `setLocalAwareness()` grows a whole `tagMap` and `publishTag()` sends that entire object as one `element:<tag>` update at lines 60-63 / 111-113. That update goes through `RealtimePresenceTransport.update()` and `validatePresenceClientMessage()` before any socket send, and the presence protocol caps each update value at 4096 bytes (`MAX_PRESENCE_VALUE_BYTES`, enforced by `assertJsonSize(value)`). Built-ins seed awareness for every matching element, e.g. `can-hover` has `myDefaultAwareness: { hover: false }` and `can-mirror` has `{ hover: false, focus: false }`, so a page with enough hover/mirror elements can exceed 4 KB just from the accumulated `{ [elementId]: awareness }` object. Once that happens, `transport.update()` throws before sending, `publishTag()` only logs a warning, and remote clients keep the previous/truncated tag map, so later elements' hover/focus/presence never sync. Publish per element, shard the tag state under bounded channels, or otherwise keep each presence value under the protocol cap instead of sending an unbounded per-tag map.

spencerc99:
## Code review

- **P1 [`packages/playhtml/src/index.ts:1236`](https://github.com/spencerc99/playhtml/blob/1dcd348fd4db5955354fc8a9ce183240acb7231e/packages/playhtml/src/index.ts#L1236) drops local awareness for elements that stay mounted across a page-room navigation.** `runHandleNavigation()` deliberately keeps connected handlers, but this teardown discards the old client and its `localTags`, then line 1253 builds an empty client. When `setupElements()` scans the still-mounted nodes, the existing-handler branch at [`index.ts:2337-2353`](https://github.com/spencerc99/playhtml/blob/1dcd348fd4db5955354fc8a9ce183240acb7231e/packages/playhtml/src/index.ts#L2337) only reinitializes callbacks and returns; it never republishes the handler's existing `selfAwareness`. Stable React/SPA nodes therefore disappear from element awareness in the new room until some later user action changes their awareness. Transfer retained handlers' current awareness into the replacement client, or explicitly republish it during reinitialization.

- **P1 [`packages/playhtml/src/element-awareness.ts:72`](https://github.com/spencerc99/playhtml/blob/1dcd348fd4db5955354fc8a9ce183240acb7231e/packages/playhtml/src/element-awareness.ts#L72) publishes synchronously for every initialized awareness-bearing element, exceeding the server's 45 Hz interactive budget on ordinary high-count pages.** Each initial `setMyAwareness` calls `setLocalAwareness`, and `publishLocalAwareness()` resends every current shard. The existing server classifies every `element:*` update as interactive and accepts only 45 per connection per second ([`partykit/presencePolicy.ts:14`](https://github.com/spencerc99/playhtml/blob/1dcd348fd4db5955354fc8a9ce183240acb7231e/partykit/presencePolicy.ts#L14)); rejected updates are not committed. This client explicitly ignores `presence-rate` at lines 177-178 and has no retry, so later elements stay absent remotely even though the local fake-socket test sees every send. Coalesce setup writes into a bounded publish and retry the latest snapshot after rate limiting.

- **P1 [`packages/playhtml/src/element-awareness.ts:124`](https://github.com/spencerc99/playhtml/blob/1dcd348fd4db5955354fc8a9ce183240acb7231e/packages/playhtml/src/element-awareness.ts#L124) creates an unbounded number of shard channels even though the server allows only 32 total channels per connection.** `identity` and `page` already consume two slots (and a shared cursor socket consumes more), so 31 near-limit awareness shards are enough for `assertPresenceChannelCapacity()` to reject the tail with `Presence channel limit exceeded` ([`partykit/presencePolicy.ts:318-324`](https://github.com/spencerc99/playhtml/blob/1dcd348fd4db5955354fc8a9ce183240acb7231e/partykit/presencePolicy.ts#L318)). Bounding each shard to 4 KB fixes the earlier value-size failure but not this aggregate channel limit; cap or multiplex the shard set within the remaining connection budget.

spencerc99:
This PR now covers BOTH halves of #240 (bundled per discussion so the whole migration reviews, tests, and releases as one unit):

**Element awareness** (original scope): `setMyAwareness` / element awareness over page-scoped presence channels via a refcounted per-room transport registry; Yjs fallback preserved.

**Page presence** (new commits `681481ab` + `17610a8d`): `playhtml.presence` custom channels publish under an internal `presence:` wire prefix (classified into the server's low-frequency event bucket — zero server changes); `createPresenceRoom` swaps its dedicated Y.Doc awareness-bus socket for an isolated presence-transport socket (public API unchanged — extension call sites untouched); no-WebSocket fallback for both preserved.

**Identity consolidation**: identity broadcasting now lives in exactly one place per socket — `acquirePresenceTransport` subscribes once to `users.onSelfChange` and re-joins on change. Per-consumer identity wiring (element awareness refresh, cursor-client transport republish) deleted; cursor-client keeps only its rendering reactions.

**Verification**: playhtml 446/446 (+29 new tests), common 16/16, react green, tsc clean, lint = pre-existing extension errors only, smoke 23/23 incl. two live cross-browser element-awareness tests.

**Review notes**: (1) in transport mode `playhtml.presence` is reassigned on page-room navigation — consumers holding a pre-navigation reference get a dead client (same semantics as element awareness; extension re-wires per page); (2) server caps 32 channels/connection — current consumers sit well under it.

**Release plan**: hold the changesets release PR until this merges, then one npm release covers the whole migration (single mixed-version window).

spencerc99:
Review response (commits `d3f97a9f`..`3b8278d5`) — all four codex P1s addressed:

1. **Per-tag map exceeding the 4KB value cap** — already fixed by shard channels (`element:shard:N`, each 4KB-bounded) before this response; no further change.
2. **Init burst vs 45Hz interactive budget** — local writes now coalesce into ONE shard-set publish per tick (microtask), and a ~400ms trailing re-publish re-sends the latest snapshot so anything the server rate-dropped recovers. Integration test: 100-element sweep produces a bounded publish count with full final state.
3. **Navigation dropping retained elements' awareness** — the rebuilt client is seeded from every retained handler's current selfAwareness in one batched publish; still-mounted SPA elements stay visible with no user action.
4. **Shard-count vs 32-channel cap** — element shards capped at 8 (~32KB); overflow drops with a single loud console.error naming the affected tags, never throws.

Also added the **stable presence facade**: `playhtml.presence` is now one object for the instance lifetime — captured references and pre-navigation subscriptions keep working across page-room changes (old Yjs-path semantics restored).

Verification at head: playhtml 462/462 (+16 new tests), tsc clean, smoke 23/23 incl. both live cross-browser element-awareness tests.

spencerc99:
Final architecture pass (commits `d40a6a0c`..`66cce0eb`) — the branch is now FROZEN pending review/merge.

**PeerStore consolidation** (per discussion — final shape now, not a later refactor): `RealtimePresenceTransport` owns one `PeerStore` per socket, the single consumer of presence-sync/changes. It folds all channels into a publicKey-keyed peer map once and notifies per-namespace (`cursor`/`element`/`presence`/`identity`) — a cursor-only update provably does not fire element/presence subscribers (unit-tested). Cursors, element awareness, and page presence are now thin subscriptions over it; `createPresenceRoom` sockets get their own store through the same mechanism. Receive-side logic net −95 LOC across the consumers, replaced by one 150-LOC store. Publish side, wire formats, and public APIs untouched — old/new clients interop unchanged.

**Pacing deletion: deliberately NOT done.** Investigation showed client cursor pacing is not a duplicate of the server's rate signal: it implements room-size fanout backoff (60→30Hz past 30 cursors) that the server's flat 90Hz budget cannot produce, and `presence-rate` is reactive-only. They're complementary; removing it would double busy-room traffic. Audit doc corrected.

**Orphan sweep**: inlined one dead wrapper, unexported 3 internal-only types; per-symbol caller count found no other dead code.

**Final state**: 471 tests green (49 files), tsc clean, smoke 23/23 incl. both live cross-browser tests, lint clean in packages/. 23 commits, +3368/−188 in packages/playhtml/src.

packages/playhtml/src/presence-client.ts
```diff
@@ -0,0 +1,304 @@
...
+}
+
+function jsonByteLength(value: unknown): number {
+  try {
+    const json = JSON.stringify(value);
+    if (json === undefined) return Infinity;
+    return new TextEncoder().encode(json).byteLength;
+  } catch {
+    return Infinity;
+  }
```
spencerc99: these common utils should not be here please put them in a common utils file and we are duplicating some of these across many of the diff files.

spencerc99:
Re: the duplicated-utils review comment — addressed in `d0f79a9c`.

All shared helpers now live once in the internal, dependency-free `presence-utils.ts` (imports only `@playhtml/common`; nothing new exported from the package): `jsonByteLength`, the `presence:` and `element:` channel prefixes (the `presence:` duplication existed to dodge a circular import — the leaf module resolves the cycle properly), channel build/parse helpers, stable-id derivation (`getPeerPublicKey`), and the byte-capped publish/clear pair both clients had near-copies of. Also deleted the hand-written structural mirror of the transport type in cursor-client (drift risk) in favor of the concrete type.

Deliberately NOT merged, because the variants differ subtly and silent unification would change behavior: `safeStringify` (null-coalescing semantics differ from the awareness fingerprint builders) and cursor-presence-store's stricter stable-id derivation (guarded, no connection-id fallback). Both noted in code.

Consumers net −83 LOC; the shared home is +101. 471 tests pass with behavioral assertions unchanged, tsc clean, smoke 23/23. Branch re-frozen.

spencerc99:
## Code review

- **P1 [`packages/playhtml/src/index.ts:1358`](https://github.com/spencerc99/playhtml/blob/ae85f352d25db55f91be56bef961e4787ad2d2cd/packages/playhtml/src/index.ts#L1358) reattaches existing cursor-channel presence subscriptions to the cursor client that is about to be destroyed.** `PresenceFacade.setInner()` immediately resubscribes active listeners; the new inner client delegates cursor subscriptions through the current `cursorClient` at lines 995-997. Navigation only tears down and rebuilds that cursor at lines 1361-1372, after the reattachment, so a `playhtml.presence.onPresenceChange("cursor", ...)` registered before navigation remains bound to the dead client and never receives updates from the replacement. Rebuild the cursor before swapping the presence inner, or make cursor subscriptions themselves rebind when `cursorClient` changes.

- **P1 [`packages/playhtml/src/presence-client.ts:205`](https://github.com/spencerc99/playhtml/blob/ae85f352d25db55f91be56bef961e4787ad2d2cd/packages/playhtml/src/presence-client.ts#L205) lets custom channels overwrite the system fields in `PresenceView`.** A peer can publish `setMyPresence("playerIdentity", value)`, `"cursor"`, or `"isMe"`; the loop strips the `presence:` prefix and assigns that key after initializing the trusted identity/cursor flags. `buildSelfView()` does the same at line 223. The previous awareness implementation explicitly filtered these reserved names, so this migration lets presence data spoof identity/ownership fields or replace the cursor value consumers read. Reject or skip `playerIdentity`, `cursor`, and `isMe` custom channel names.

- **P2 [`.changeset/page-presence-transport.md:5`](https://github.com/spencerc99/playhtml/blob/ae85f352d25db55f91be56bef961e4787ad2d2cd/.changeset/page-presence-transport.md#L5) truncates the release note to a bare implementation sentence.** The same happened in `element-awareness-presence-transport.md`. The repository requires each package changeset to be a user-facing paragraph explaining the change and why; the latest commits delete the navigation behavior, reliability limits, compatibility note, and rationale. Restore concise user-facing descriptions of the observable behavior and reason for both changesets.

spencerc99:
Final review + hardening batch pushed (`41981705`..`0cc8135f`). Two independent review passes (lifecycle/concurrency + silent-failure audit) over the full branch diff; every finding fixed or explicitly documented:

**Blockers fixed:**
1. Cursor-channel `onPresenceChange` subscriptions died on navigation / server room reset (both reviewers, independently — survived tests because the presence-navigation suite ran cursors-disabled). Fixed with a stable cursor-presence hub resolving the live cursor client at dispatch time; regression tests now run cursors-ENABLED for nav + server reset.
2. No exception isolation in the dispatch chain — one throwing consumer callback skipped all remaining listeners on the shared socket. `safeInvoke` applied at every fan-out layer; `PresenceClient` commits its dedup fingerprint only on successful delivery.

**Hardening:** stale localTags purge on same-room DOM swaps; presence rate-drop trailing republish (mirrors element awareness); `presence-error`/`presence-rate` now observable on every socket (previously silent on cursors-disabled pages and all `createPresenceRoom` sockets); dead-transport observability (close/error listeners, `connectionState`, one loud error after 3 failed reconnects / 15s no-open); refcount release on constructor throw; identity re-join failures warn instead of `catch {}`.

**Documented, not changed:** multi-tab winner-selection divergence across views; `latestJoin.page` staleness on fixed-custom-room SPA nav.

**One assertion moved** (flagged, not silent): the cursor client's presence-error warn test now asserts the transport owns that warn.

**Final state: 485 tests (50 files), tsc clean, smoke 23/23, lint clean in packages/. Branch frozen for real this time.**

spencerc99:
## Code review

- **P1 [`packages/playhtml/src/presence-client.ts:153`](https://github.com/spencerc99/playhtml/blob/0cc8135f1e05139feb16ff663f07dfc34e89d164/packages/playhtml/src/presence-client.ts#L153) schedules rate-drop recovery while the server budget is still exhausted.** The server keeps a fixed 1,000 ms event window and rejects everything after 20 messages, but this retry runs only 400 ms after a short burst. For example, after 21 immediate `setMyPresence` calls, the 21st update is dropped and the trailing replay is also dropped inside the same saturated window; no later retry occurs. A dropped clear is permanent as well: line 92 deletes it from `localChannels`, the replay only iterates remaining values, and another clear returns early at line 91. Retry after the budget window (or from the server rate signal) and retain pending clears until they are replayed.

- **P1 [`packages/playhtml/src/presence-client.ts:252`](https://github.com/spencerc99/playhtml/blob/0cc8135f1e05139feb16ff663f07dfc34e89d164/packages/playhtml/src/presence-client.ts#L252) still lets custom presence channels overwrite trusted `PresenceView` fields.** `setMyPresence("playerIdentity", fake)` publishes `presence:playerIdentity`; this loop assigns that value after initializing the identity from the trusted `identity` channel, and `buildSelfView` does the same at line 270. `isMe` is equally overwriteable. Reject these reserved public channel names on write and ignore them when folding remote custom channels.

- **P2 [`.changeset/element-awareness-presence-transport.md:5`](https://github.com/spencerc99/playhtml/blob/0cc8135f1e05139feb16ff663f07dfc34e89d164/.changeset/element-awareness-presence-transport.md#L5) still does not meet the repository changeset requirement.** The package instructions require a user-facing paragraph explaining the change and why; this remains a bare transport-mechanism sentence with no user-facing reason or consequence. Expand it into the required release-note paragraph.

spencerc99:
Response to the latest codex sweep + the migration parity audit (commits `dfb69add` + `5e50331e`):

1. **Rate-drop recovery (P1)** — confirmed: the ~400ms trailing replay landed inside the server's fixed 1,000ms budget window and was itself dropped, and a dropped `clear` was permanently lost (deleted locally before send, replay only re-sent surviving values). Fixed in BOTH clients: trailing republish moved past the window (1,100ms), and pending clears are tracked and re-issued until sent. Element awareness had both defects too (its full-snapshot republish covered dropped updates but not dropped clears) — fixed identically.

2. **Reserved-field spoofing (P1)** — restored the old Yjs `SYSTEM_FIELDS` guard lost in migration, on both ends: `setMyPresence` throws on `playerIdentity`/`cursor`/`isMe`, and the remote/self fold skips reserved keys (the actual security boundary against peers bypassing client validation).

3. **Changesets (P2)** — deliberately not changed; Spencer trimmed them intentionally.

4. **Parity-audit finding: unified client-side stale-peer expiry.** A full old-vs-new behavior audit found one more silent loss: Yjs's ~30s client-side stale-peer backstop survived only for cursors; element awareness and page presence trusted server close-detection completely (ghost presences on ungraceful disconnects). Now ONE shared mechanism in `PeerStore` (timestamp envelope on the wire — added pre-release while the format is still free, shared 30s constant, per-fold + 1s sweep, expiry excluded from change-fingerprints so keepalives never re-fire callbacks), with the cursor expiry REFACTORED ONTO it — one place in the codebase now defines staleness. Cursor behavior preserved (61 cursor tests, assertions unchanged); deliberate asymmetry documented in code: cursors fade to identity-only with no keepalive, element/presence keepalive quietly.

**State: 498 tests (50 files), tsc clean, smoke 23/23, lint clean in packages/.**

PR #280, opened 2026-07-07
wewere.online: cursor trail experiments — relay, redraw, touches

spencerc99:
## Summary

Three new visualization experiments over the extension's archived cursor data, all read-side only (no worker or extension changes), plus the shared infrastructure they ride on. Each serves at its own path on wewere.online.

### `/relay` — trail relay
Chains archive cursor trails end-to-origin into one continuous connect-the-dots journey. Every junction renders up front as a hollow numbered dot and fills with a pop as the relay cursor arrives, with a synthetic click ripple at each handoff. Chaining is greedy nearest-origin with k-nearest sampling; knobs for trail-count vs total-distance caps, per-trail length cap, sequential dot spacing (150px floor), all-pairs dot separation (30px), edge-exit filtering, and a rest beat between trails.

### `/redraw` — cursor redraw
Draws an uploaded image out of real cursor trails. Three modes judged side by side:
- **mosaic** — rigidly places the shape-nearest real gesture over each traced image contour (translate/scale only, never bent)
- **warp** — pairs gestures to contours by length rank and bends each point toward its arc-position twin, with a strength dial from raw gestures to exact trace
- **in place** — nothing moves: trails stay at their true positions and only the segments passing through the image's edge corridor are revealed, so the image emerges from where people actually were (attention bias becomes part of the drawing)

### `/touches` — cursor touches (p5)
Replays trails on their real timestamps — compressed to co-presence windows — and marks every moment two cursors from different pages touched at the same real time. Touch detection requires both cursors genuinely in motion (tight sample brackets), which took the count from 45k phantom touches to ~2k real ones. Live cursors read as koi: small glowing bodies, thin wakes, a before-glow tracing both paths toward each upcoming meeting. On contact, a large clasp of the two hand cursors from experiment 7 lands in the pair's colors, holds for a beat inside a stacked click ripple (geometry from CLICK_DEFAULTS, fading instead of persisting), then shrinks and bleaches into a permanent sun-print-style mark — internet patina accumulating over the cycle. Linen is the default; night mode, three renderers (nebula 2d / glsl webgl / original pinwheel), and five mark styles are switchable for comparison.

## Shared infrastructure
- `useCursorEventPool` — first 20k events render immediately, older pages auto-fetch in the background up to 100k (~4k trails)
- `useChromeToggle` — double-press `d` hides page chrome, mirroring the movement page's shortcut
- `trailSequence` — path-length-paced sequential/overlapping scheduling for `AnimatedTrails`
- `p5` added as a website dependency (touches page only)

## Notes
- No changeset: nothing under `packages/` changed (the `bun.lock` commit is a version sync for the already-published playhtml 2.11.3).
- No docs/starter-template changes needed: no public core API touched.
- Verified live against production archive data throughout (screenshots in session); happy to record proper footage for the changelog before release.


spencerc99:
## Code review

- **P2 [`extension/PENDING.md:4`](https://github.com/spencerc99/playhtml/blob/c1dc7bd0b9c213e26ac43c3ccb8e4189c32827cd/extension/PENDING.md#L4) narrows the extension release-note rule in a way that conflicts with the repo instructions for this PR.** The root project instructions require a short `extension/PENDING.md` bullet whenever a PR touches `extension/**` or changes extension-facing behavior, but this PR changes the pending-note guidance to require bullets only for `extension/src/**` / zip-shipped changes and explicitly exempts `extension/website/**` at lines 4-6. This same PR adds multiple new `extension/website/**` experiments, so under the current repo instructions it still needs a pending release-note bullet instead of weakening the rule. Restore the broader rule or get Spencer to change the project instruction first, then add the appropriate bullet for these website experiments.

PR #189, opened 2026-06-18
inventory: the shared satchel (system tools, no economy yet)

spencerc99:
## Summary

Adds **inventory** — the shared layer both object-mechanics (bottles, quarantine-tape) pull items from. It governs holding, equipping, and (later) acquiring the things you use as you browse. Today both features assume their item is always available; this is the layer underneath that actually governs what you hold and how you wield it.

**v1 scope: system-tools-only, no economy.** Permanent tools (the bottle-sealer, eventually the tape-gun) are infinite, so bottles/tape ship fully usable with no economy code — that part is deliberately deferred and decided per-item later.

## What's in it

- **Pure core** (unit-tested): `registry`, `armed-state`, `held-store`, `InventoryManager` → the `InventoryAPI` (`register / list / arm / disarm / getArmed / onArmedChange / count`).
- **Surface**: a draggable backpack **edge-nub** that opens the cozy textured **Satchel** (Shadow-DOM injected, warm-linen aesthetic). Click-to-arm; the armed item **wields beside your cursor** (cursor stays visible). `Esc` disarms.
- **Keyboard summon**: reliable **manifest `commands`** (`Cmd+Shift+I` mac / `Ctrl+Shift+E` win-linux) opens the kit at the cursor — browser-routed so it works on every site, unlike a content-script keydown.
- **Integration**: `GlobalFeatureDeps` gains `inventory`; the manager is built **before** the experiment loop so `deps.inventory` exists when each experiment registers. `bottles` registers a `message-bottle` item. `INVENTORY` flag added. `inventory/*` exposed via `web_accessible_resources`.

## Why base = `feat/social-bottles`

Inventory extends the social-experiment scaffold (`features/social/`, `global.ts`) which lives on the bottles branch, not `main`. This branch was rebased onto `origin/feat/social-bottles`, so the diff here is **inventory-only** (~19 commits). Merge order: bottles → inventory → quarantine-tape rebases on inventory.

## Test plan

- [x] All extension unit tests pass (281).
- [x] Extension builds clean (`bun run build`); inventory assets ship; manifest has the command + `inventory/*` resources.
- [x] Playwright recording of the full flow against the **real built extension** (drag nub → open kit → arm → wield → Esc).
- [ ] **In-browser human verification** — drag/arm/wield/keyboard on a real page (the OS-cursor-beside-item read can't be captured in a Playwright video).

## Before merge (tracked)

[...]

PR #281, opened 2026-07-07
sealing: textured outer roll, seal-band beat, fix finale/band alignment

spencerc99:
## Summary

Polish + fixes for the message-bottle sealing ceremony (follow-up to #189):

- **Textured outer roll** — the rolled-up message now shows the message texture and the author color stripe on its outer face (back material renders the shared canvas texture with a light grey tint), so the rolled card matches the on-page capsule instead of reading as blank white paper.
- **Seal-band beat** — a new animation beat: after the roll fits to card size, a band in the author's color snaps around the coil (spring `scaleX` sweep) like sealing an envelope, then fades as the bottle lifts into the finale.
- **Alignment fixes** — the finale's world↔screen mapping ignored `camera.zoom` (0.65 after the zoom-out), so the coil only traveled 65% of the way to the slot; and the coil winds around the paper's top edge, so it sits offset from the mesh origin. `playFinale` now converts slot coordinates through the zoom and corrects by the coil's bounding-box center, and `playSealBand` projects the coil bbox to screen the same way.

## Testing

- Verified with trusted-input Playwright runs on the social playground (`/social-playground/`): coil landing measured pixel-exact against the slot (`coil@screen {x:752, bottomY:432}` === slot `{752, 432}`), band rendered centered on the coil.
- Clean end-to-end ceremony run with zero console errors/warnings after removing debug instrumentation.
- `tsc` clean for the sealing files (remaining workspace errors are pre-existing and untouched).

No changeset: changes are under `extension/`, not `packages/`. No docs change: internal extension behavior, not library API. No `PENDING.md` bullet: bottles are behind a flag and not public yet.

### DIFF
--- a/extension/PENDING.md
+++ b/extension/PENDING.md
@@ -14,3 +14,5 @@ extension/website/public/changelog/media/ and reference them here:
 ![Screenshot title](/changelog/media/file.png)
 ![video: Demo title](/changelog/media/file.mp4)
 -->
+
+- Sealing a message bottle got a nicer send-off: the rolled-up message keeps its text and your color on the outside, a color band snaps around it to seal it like an envelope, and it now drops exactly into the slot in the page.
--- a/extension/PENDING.md
+++ b/extension/PENDING.md
@@ -14,5 +14,3 @@ extension/website/public/changelog/media/ and reference them here:
 ![Screenshot title](/changelog/media/file.png)
 ![video: Demo title](/changelog/media/file.mp4)
 -->
-
-- Sealing a message bottle got a nicer send-off: the rolled-up message keeps its text and your color on the outside, a color band snaps around it to seal it like an envelope, and it now drops exactly into the slot in the page.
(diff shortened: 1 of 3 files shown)


spencerc99:
## Code review

- **P2 [`extension/src/components/sealing/SealingCeremony.tsx:68`](https://github.com/spencerc99/playhtml/blob/b0f8cd1c5afa4ec58583ae41b9eacb08be7eb18e/extension/src/components/sealing/SealingCeremony.tsx#L68) changes extension-shipped behavior without the required pending release-note bullet.** This PR changes `extension/src/**` sealing behavior, but `extension/PENDING.md` remains unchanged even though its repo instruction says to add a bullet for any PR that touches `extension/**` (`extension/PENDING.md:4` in this head). The extension release-prep flow only moves bullets from `PENDING.md` into the changelog, so this user-facing sealing polish would ship without release-note coverage. Add a short bullet to `extension/PENDING.md`, or move the change out of the shipped extension surface if it should not be part of the extension release.

PR #283, opened 2026-07-07
Coalesce can-mirror observer updates

spencerc99:
## Summary

- Coalesce `can-mirror` DOM observer writes before syncing so many simultaneous element mutations flush as one Yjs transaction per frame.
- Keep direct handler `setData` immediate so explicit programmatic updates still apply synchronously.
- Add a network-pacing regression that verifies many mirrored pixel mutations produce one outbound document update instead of one per element.

## Root Cause

The skywriting example has 2,000 `can-mirror` elements and a 100ms fade loop over `.writing` pixels. Each mirrored pixel owned its own observer, and each observer write became its own Yjs document update. That could exceed the server's per-connection websocket message rate before the server could do useful batching.

## Validation

- Fetched the live GitHub Pages example and confirmed it matches the pasted 40x50 pixel reproduction.
- `bun run test` in `packages/playhtml` (350 tests)

No docs or starter-template changes: this is runtime pacing only and does not change public API or usage.

### DIFF
--- a/.changeset/can-mirror-pacing.md
+++ b/.changeset/can-mirror-pacing.md
@@ -1,0 +1,5 @@
+---
+"playhtml": patch
+---
+
+Coalesce can-mirror DOM observer writes before syncing so many simultaneous element mutations send a single shared document update instead of one update per element.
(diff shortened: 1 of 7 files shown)


spencerc99:
## Code review

PR #284, opened 2026-07-07
Add playhtml.users identity API; consolidate identity ownership

spencerc99:
## Summary

Adds a core `playhtml.users` module owning durable user identity (name and color) that works whether or not cursors are enabled, and folds in the identity-consolidation simplifications from the 2026-07-24 audit (Bucket 2). Conceptual split: **users** = who people are (durable identity), **presence** = what they're doing right now (live channels), **cursors** = the rendered layer.

### API

- `playhtml.users.me` — `{ pid, name, color }`; setters persist to localStorage and publish to the room.
- `playhtml.users.getAll(): User[]` / `onChange(cb)` — everyone in the room (`{ pid, name, color, isMe }`), merging main-room identities with cursor-room presences.
- React: new `useUsers(): User[]` hook; `usePlayerIdentity()` no longer requires `cursors: { enabled: true }`.
- New top-level `InitOptions.playerIdentity` (`cursors.playerIdentity` still honored, takes precedence).
- `window.cursors.color/name` are thin delegates to `users.me`; `allColors` is derived from the room's users.

Custom per-user properties (`me.custom` / `setCustom`) were built on this branch and then deliberately deferred to a later release; nothing custom-related ships here.

### Simplifications (audit Bucket 2, ~590 net lines removed)

- Identity stored and saved in ONE place (previously three stores kept in sync); identity changes now republish to peers — fixing a real bug where a name/color change on a cursor-less page never reached other visitors.
- Extension identity bridge targets `users.adoptIdentity()` directly and works without cursors (also resolves the earlier codex review P1).
- `resolveMyIdentity()` replaces four copies of the identity fallback expression.
- `allColors` aggregation moved out of the cursor renderer into a users selector; the undocumented `window.cursors.allColors` setter is removed.
- React: one internal `usePlayhtmlSubscription` primitive backs the five subscription hooks; `PlayProvider` no longer hand-mirrors cursor identity state.
- React migrated off the raw `elementHandlers` map onto `getHandle` (its first real consumer); `elementHandlers` is no longer on the public `playhtml` singleton.

### Breaking/surface notes

- `playhtml.elementHandlers` removed from the public singleton (use `getHandle(elementId, tag)`); internal named export remains for package tests.
- `window.cursors.allColors` setter removed (getter unchanged; was undocumented and internal-only).

## Verification

playhtml 364/364, react 44/44, common 19/19, partykit 105/105; `tsc` exit 0 in common/playhtml/react and `tsc -p partykit`. Docs (users page, cursors page, reference pages) and the changeset rewritten to match the shipped surface; repo-wide grep confirms no custom-properties references remain in code or docs.

Starter templates unchanged: additive API only; existing starter usage stays valid.


spencerc99:
## Code review

- **P1 [`packages/playhtml/src/index.ts:1355`](https://github.com/spencerc99/playhtml/blob/3c974ef61148315cb8fdd4fa54677c943f9d6ffe/packages/playhtml/src/index.ts#L1355) keeps the extension identity bridge behind `cursors.enabled`, so cursor-disabled pages never adopt the extension identity.** This PR creates `usersAPI` unconditionally at lines 1337-1345 and documents that `playhtml.users` / `usePlayerIdentity()` work without cursors, but the `playhtml:configure-identity` listener is only installed when cursors are enabled at lines 1355-1357. Even if this listener were called on a cursor-disabled page, the handler immediately returns when `cursorClient` is absent at lines 1096-1097 and applies the merge through `cursorClient.configure()` at line 1106. A `PlayProvider` or `playhtml.init()` without `cursors: { enabled: true }` will therefore keep the generated/default PID and color instead of the extension-injected PID/color, and `playhtml:ready` is not dispatched for the extension path. Install the listener after `usersAPI` is created regardless of cursor settings, and route the merged identity through `usersAPI.adoptIdentity(...)` when there is no cursor client.

spencerc99:
## Code review

- **[P1] Migrate the docs runtime before removing `elementHandlers`.** The PR removes the property from the public singleton in [`packages/playhtml/src/index.ts`](https://github.com/spencerc99/playhtml/blob/8f905e310ff091da4db82922bbc87f8b90e5cd12/packages/playhtml/src/index.ts#L2034-L2059), but the unchanged docs client still reads it for the code-block handler and for two readiness polls in [`apps/docs/src/scripts/enhance-code-blocks.ts`](https://github.com/spencerc99/playhtml/blob/8f905e310ff091da4db82922bbc87f8b90e5cd12/apps/docs/src/scripts/enhance-code-blocks.ts#L309-L318) and [`apps/docs/src/scripts/enhance-code-blocks.ts`](https://github.com/spencerc99/playhtml/blob/8f905e310ff091da4db82922bbc87f8b90e5cd12/apps/docs/src/scripts/enhance-code-blocks.ts#L346-L367). With this head, both `Boolean(playhtml.elementHandlers)` polls remain false forever, so code-block wear handlers and the shared copy-event listener never register; the click path also has no handler to update. A targeted TypeScript check reports TS2551 at lines 309, 357, and 599. Replace the readiness probe with the supported lifecycle API (`playhtml.ready`/`isLoading`) and use `getHandle(elementId, "can-play").getData()/setData()`.

- **[P1] Rebind the users roster to a replacement cursor client.** [`createUsersAPI`](https://github.com/spencerc99/playhtml/blob/8f905e310ff091da4db82922bbc87f8b90e5cd12/packages/playhtml/src/users.ts#L167-L177) saves the first `onCursorPresencesChange` unsubscribe and never clears it until `destroy()`. Navigation destroys and rebuilds `cursorClient`, then only calls `usersAPI.getAll()` ([`packages/playhtml/src/index.ts`](https://github.com/spencerc99/playhtml/blob/8f905e310ff091da4db82922bbc87f8b90e5cd12/packages/playhtml/src/index.ts#L1189-L1204)); because the stored unsubscribe remains truthy, `ensureSubscribed()` never attaches to the replacement client. After a route/cursor-room change, `useUsers()` and derived `allColors` stop reacting to cursor-room joins, leaves, and identity changes unless an unrelated main-awareness event happens. A focused test that swaps the dependency's cursor source and calls `getAll()` reproduces this: the replacement callback is never registered. Make the cursor subscription source-aware, or explicitly unsubscribe/rebind it during cursor teardown/rebuild.

- **[P1] Release the public API removal as a major change.** [`.changeset/users-api.md`](https://github.com/spencerc99/playhtml/blob/8f905e310ff091da4db82922bbc87f8b90e5cd12/.changeset/users-api.md#L1-L9) marks `playhtml` as `minor` while explicitly removing the documented `playhtml.elementHandlers` property from stable `playhtml@2.13.1`. Existing consumers compiled against that property receive `undefined` after a minor upgrade. If the removal is intentional, mark the `playhtml` changeset as `major`; otherwise retain the property.

PR #285, opened 2026-07-07
Allow awareness-only can-play elements

spencerc99:
## Summary

- allow `can-play` initializers to be valid when they render only element awareness via `updateElementAwareness`
- make initializer diagnostics specific about incomplete `defaultData` and `myDefaultAwareness` pairs
- reject data render paths (`updateElement` / `view`) that omit `defaultData`
- reject `null` / primitive `defaultData` in setup-time validation
- update the public element API docs and awareness examples to avoid placeholder shared data for awareness-only elements
- add changesets for `playhtml`, `@playhtml/common`, and `@playhtml/react`

## Root Cause

The setup guard always required `defaultData` plus `updateElement` or `view`, so an element that only declared `myDefaultAwareness` and `updateElementAwareness` was rejected even though awareness rendering is a valid update path. The validation also needed to report both sides of the data pair: `defaultData` without a data render path, and a data render path without `defaultData`.

## Validation

- `packages/common`: `bun run build`
- `packages/playhtml`: focused initializer validation test
- `packages/playhtml`: `bun run test` (34 files / 351 tests)
- `packages/playhtml`: `bun run build`
- `packages/react`: `bun run test` (6 files / 43 tests)
- `packages/react`: `bun run build`
- `apps/docs`: `bun run build`
- `git diff --check`

## Review Follow-up

Addressed Spencer's review note that data update functions without `defaultData` were not being reported as invalid. Then expanded setup-time coverage across empty, data-only, render-only, awareness-only, mixed invalid-pair, primitive default, and null default cases.

### DIFF
--- a/.changeset/calm-awareness-pairs.md
+++ b/.changeset/calm-awareness-pairs.md
@@ -1,0 +1,7 @@
+---
+"playhtml": patch
+"@playhtml/common": patch
+"@playhtml/react": patch
+---
+
+Allow `can-play` elements to render from element awareness alone when `myDefaultAwareness` is paired with `updateElementAwareness`, and report incomplete initializer pairs with specific diagnostics.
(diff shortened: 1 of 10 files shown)


spencerc99:
## Code review

PR #286, opened 2026-07-08
Enforce public identity boundary for presence

spencerc99:
## Summary

- Store extension identity as `{ public, privateKey }` and keep discovered sites in separate profile storage.
- Expose public identity through explicit background messages; content initialization and `EventBuffer` no longer read the combined private storage record.
- Centralize public `PlayerIdentity` projection in `@playhtml/common`, with field allowlists, bounded strings, and bounded color palettes.
- Remove the generic 4 KiB presence-value limit and the 8 KiB raw-message limit.
- Keep a 1 MiB pre-parse server abuse ceiling, then use Cloudflare's actual hibernation attachment write as the cumulative connection-state boundary.
- Persist candidate connection state before mutating broadcast state, restoring the prior attachment if the platform rejects it.
- Snapshot validated client presence state so rejected or caller-mutated values cannot be replayed on socket open or reconnect.
- Consolidate extension identity/color storage ownership and remove stale participant/color helper paths.

## Root cause

The extension stored public identity, private signing material, and browsing-derived profile fields in one flat `playerIdentity` object. Page-facing code could receive or load that full object before selecting public fields, producing oversized Wikipedia presence payloads and crossing the intended private/public boundary.

The final transport audit also found that `join()` and `update()` cached caller-owned values before validation. A rejected value could remain queued and throw again during `flushCurrentState()`, while an accepted object could be mutated after validation.

## Security boundary

- `playerIdentity.public` is the only identity shape returned by identity/profile messages, DOM identity events, and presence.
- `playerIdentity.privateKey` remains extension-local and is not returned by those APIs.
- Content-script identity consumers request public identity from the background instead of reading the combined storage record.
- Discovered sites remain separate profile data and are not included in public identity or presence.
- Presence join and update paths reject fields outside the public identity schema.

Moving signing material from content-script-readable `browser.storage.local` into a background-owned non-extractable key store remains part of the signing design in #264. This PR fixes accidental data flow without redesigning signing.

## Presence limits

Shared presence validation now checks JSON serialization and known-channel schemas without imposing a generic byte threshold. PartyKit rejects raw messages above 1 MiB before parsing as an abuse safeguard.

Presence channels must survive Durable Object hibernation, so cumulative connection state is persisted through `connection.setState()`. The server derives and validates candidate channels, attempts the real platform attachment write, and only then updates room state. A rejected attachment restores the prior accepted state and returns a presence error without broadcasting the candidate.

## Review

Final review covered the complete branch diff, current `origin/main`, runtime-message and DOM boundaries, content-script storage access, persisted identity migration, presence join/update/reconnect paths, hibernation attachment atomicity, extension profile access, docs, release notes, and project instructions.

An independent final review found the `EventBuffer` content-script storage read; that path was moved behind `GET_PUBLIC_PLAYER_IDENTITY` and re-reviewed with no remaining high-confidence findings.

## Validation

- Extension: 53 files, 333 tests passed
- PlayHTML: 34 files, 355 tests passed
- Common presence protocol: 19 tests passed
- PartyKit presence policy/server: 19 tests passed
- PartyKit TypeScript check passed
- Common, PlayHTML, and extension production builds passed
- Combined website/docs build passed
- `git diff --check` passed
- `bun run lint` still has the previously documented unrelated extension visualization, website shared, worker Cloudflare-global, and WXT manifest type errors

Residual risk: Cloudflare attachment rejection is covered by atomic persistence/rollback unit tests but was not forced in a live Durable Object runtime.


extension/src/utils/presenceIdentity.ts
```diff
@@ -0,0 +1,49 @@
+// ABOUTME: Builds the small player identity object shared through realtime presence.
+// ABOUTME: Excludes extension-local profile and key material from page-facing payloads.
+
+import type { PlayerIdentity } from "@playhtml/common";
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return value !== null && typeof value === "object" && !Array.isArray(value);
+}
```
spencerc99: stop re-doing these simple helper functions and pull them from a common place

extension/src/storage/playerIdentity.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+function toPrivateKey(value: unknown): JsonWebKey | null {
+  if (!isRecord(value)) return null;
+  return value as JsonWebKey;
+}
+
+function hasOnlyKeys(value: Record<string, unknown>, keys: string[]): boolean {
+  return Object.keys(value).every((key) => keys.includes(key));
+}
```
spencerc99: these all seem like not super useful helper functions and redundat with other places..

packages/common/src/cursor-types.ts
```diff
@@ -14,8 +14,6 @@ export type PlayerIdentity = {
     colorPalette: string[];
     cursorStyle?: string;
   };
-  discoveredSites?: string[];
-  createdAt?: number;
```
spencerc99, line 18: this is probably still useful to keep

packages/playhtml/src/index.ts
```diff
@@ -1103,6 +1094,28 @@ function setupExtensionIdentityListener(): void {
   document.dispatchEvent(new CustomEvent("playhtml:ready"));
 }
 
+function mergeExtensionPlayerIdentity(
```
spencerc99: why do we need this? cant we just do logic like we did before?

spencerc99:
## Code review

extension/src/storage/EventBuffer.ts
```diff
@@ -28,7 +28,11 @@ function shouldStoreWithoutDelay(event: CollectionEvent): boolean {
 
 async function getEventParticipantPublicKey(): Promise<string> {
   try {
-    const identity = await getPublicPlayerIdentity();
+    const identity = toPublicPlayerIdentity(
+      await browser.runtime.sendMessage({
+        type: 'GET_PUBLIC_PLAYER_IDENTITY',
+      }),
+    );
```
spencerc99: why did we replace this?

partykit/presenceMessage.ts
```diff
@@ -0,0 +1,23 @@
+// ABOUTME: Validates incoming presence messages before the server parses them.
+// ABOUTME: Restores hibernation state when platform attachment writes are rejected.
+
+const MAX_PRESENCE_MESSAGE_BYTES = 1024 * 1024;
+
+export function assertPresenceMessageSize(message: string): void {
+  if (new TextEncoder().encode(message).byteLength > MAX_PRESENCE_MESSAGE_BYTES) {
+    throw new Error("Presence messages must be 1 MiB or less");
+  }
+}
```
spencerc99: i thought wse were just letting the server limitation handle why do we need our own?

extension/src/utils/presenceIdentity.ts
spencerc99: Done. I removed the extension-local presence identity helper and centralized the projection in `@playhtml/common` as `toPublicPlayerIdentity`, which is now shared by the extension and core library.

extension/src/storage/playerIdentity.ts
spencerc99: Done. I removed the redundant record/string/private-key helper layer and consolidated parsing around the shared `isPresenceRecord` and `toPublicPlayerIdentity` functions. The remaining functions own distinct storage, migration, key generation, and profile operations.

packages/common/src/cursor-types.ts
spencerc99: Agreed. `createdAt` is still part of `PlayerIdentity`; only `discoveredSites` moved out because browsing history is profile-local data and should not be included automatically in public presence.

packages/playhtml/src/index.ts
spencerc99: Simplified. The separate merge helper is gone. The listener now validates the incoming public identity and configures it directly, while preserving the page-owned display name.

extension/src/storage/EventBuffer.ts
spencerc99: This replacement enforces the private/public boundary. `getPublicPlayerIdentity()` lives in the storage module and reads the combined `{ public, privateKey }` record. `EventBuffer` can run in a content script, so it now asks the background process for the public projection only and never reads the secret-bearing storage object. The resulting event `pid` is still the same public key.

partykit/presenceMessage.ts
spencerc99: Agreed. This was still an application-defined 1 MiB ceiling, so I removed it in `6e284a8e`. Incoming message size now relies on the WebSocket runtime constraint; cumulative presence state relies on the actual attachment write, with rollback when the platform rejects it.

PR #292, opened 2026-07-10
Fix Firefox session coordination and packaging compatibility

spencerc99:
## Summary

- coordinate one browser-session ID in the background and serve it to content scripts over runtime messaging
- remove the Chromium-only `storage.session.setAccessLevel` dependency and the Firefox webpage-console warning
- declare `options_ui.open_in_tab` through WXT entrypoint metadata and validate it in generated manifests
- document retained Firefox compatibility branches with direct Mozilla/MDN links

## Root cause

Firefox does not implement `StorageArea.setAccessLevel`, so content scripts could not access `browser.storage.session`. Each document fell back to its own in-memory session ID and logged a warning into the inspected page's console.

The compatibility audit also found that WXT reconstructs `options_ui` from the HTML entrypoint, which caused the value in `wxt.config.ts` to be dropped from generated manifests.

## User impact

Firefox now uses one stable session identity across tabs and navigations without adding warnings to webpage developer consoles. The setup page is also consistently opened in a dedicated tab in both generated browser manifests.

## Validation

- `bun run test -- --run`: 50 files, 334 tests passed
- Chrome MV3 production build passed
- Firefox MV2 production build and zip passed
- Chrome and Firefox generated-manifest validation passed
- `web-ext lint`: 0 errors, 17 pre-existing generated-bundle warnings
- Firefox 152.0.1 live smoke: temporary add-on installed, content page initialized, and the page console contained no `ConsoleAPI` or `PageError` messages
- `git diff --check cd1ebf95..HEAD` passed

The repository-wide type-check still reports unrelated existing errors in extension visualization/worker files and the WXT schema; no changed-code type error remains.

### DIFF
--- a/extension/PENDING.md
+++ b/extension/PENDING.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - The inventory shortcut moved to Ctrl+Shift+B on Windows/Linux (Cmd+Shift+I on Mac is unchanged), freeing Ctrl+Shift+E for an in-progress emote wheel.
+- Firefox now keeps one reliable browser-session identity without adding warnings to webpage consoles.
 
 <!--
 Add a bullet here in any PR that changes the extension itself (extension/src/**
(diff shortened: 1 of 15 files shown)


spencerc99:
## Code review

PR #299, opened 2026-07-12
worker: keep metadata history monotonic

spencerc99:
## Summary

- Keep page metadata history monotonic by observation timestamp.
- Ignore delayed or duplicate navigation snapshots that are not newer than the current row.
- Cover older retries, exact duplicate event IDs, and chronological handling of an out-of-order batch.

## Root cause

The history write ran after the event upsert but queried only the current metadata hash. A delayed retry with different metadata could close a newer row at an older timestamp and replace it as current.

## Validation

- `bun run test` in `extension/worker` (45 tests)
- Conservative local review: no findings.

### DIFF
--- a/extension/db/migrations/page_metadata_history_atomic.sql
+++ b/extension/db/migrations/page_metadata_history_atomic.sql
@@ -1,3 +1,6 @@
+-- ABOUTME: Serializes page metadata history transitions in PostgreSQL.
+-- ABOUTME: Enforces one current row and tracks the latest observation timestamp.
+
 do $$
 begin
   if exists (
@@ -18,6 +21,16 @@ create unique index idx_page_metadata_history_current
   on public.page_metadata_history (page_ref)
   where valid_to_ts is null;
 
+alter table public.page_metadata_history
+  add column if not exists latest_observed_at_ts timestamptz;
+
+update public.page_metadata_history
+set latest_observed_at_ts = valid_from_ts
+where latest_observed_at_ts is null;
+
+alter table public.page_metadata_history
+  alter column latest_observed_at_ts set not null;
+
 create or replace function public.record_page_metadata_snapshot(
   p_page_ref text,
   p_canonical_url text,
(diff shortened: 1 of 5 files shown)


spencerc99:
## Code review

- **P1 [`extension/db/migrations/page_metadata_history_atomic.sql:46`](https://github.com/spencerc99/playhtml/blob/f02a6b192122fbc8bbf79c0755902caf3ed64a5a/extension/db/migrations/page_metadata_history_atomic.sql#L46-L52) still lets a stale, different snapshot replace metadata observed later.** The function compares an incoming timestamp with `valid_from_ts`, but a newer observation with the same hash returns at lines 50-51 without advancing any watermark. Starting with current `A@100`, processing `A@300` leaves the comparison timestamp at 100; a delayed `B@200` then passes the timestamp check, closes A at 200, and becomes current even though A was observed at 300. Store a separate latest-observation timestamp and advance it for same-hash observations, then reject every incoming snapshot at or below that watermark.

- **P2 [`extension/worker/src/__tests__/ingest.test.ts:189`](https://github.com/spencerc99/playhtml/blob/f02a6b192122fbc8bbf79c0755902caf3ed64a5a/extension/worker/src/__tests__/ingest.test.ts#L189-L227) tests serialization implemented by the mock instead of the SQL function.** The mock itself applies the monotonic timestamp rule, so the assertion cannot validate the advisory lock, transaction, or database behavior and conflicts with the project rule against tests of mocked behavior. Exercise concurrent calls against real PostgreSQL/Supabase, or narrow this unit test to request/RPC plumbing without claiming it verifies serialization.

- **P3 [`extension/db/migrations/page_metadata_history_atomic.sql:1`](https://github.com/spencerc99/playhtml/blob/f02a6b192122fbc8bbf79c0755902caf3ed64a5a/extension/db/migrations/page_metadata_history_atomic.sql#L1) violates the repository's mandatory two-line `ABOUTME:` header rule for code files.** Add the required two-line file description.

PR #302, opened 2026-07-12
Fix stale archive filter results

spencerc99:
## Summary

- ignore archive event responses that finish after a newer filter or refresh request begins
- add deterministic out-of-order component coverage for stale success and stale errors

## Root cause

Each request could update event state and fetch bookkeeping after it awaited its network response, even if a later filter request had already replaced the active context.

## Validation

- `bun run build-packages`
- `bun run test --run` (341 tests)
- `bun run build` in `extension/website`

`bun run lint` in `extension/website` reaches TypeScript successfully but then fails because the installed ESLint 9 toolchain cannot find a flat config.

## Integration

`origin/feat/installation-multiscreen` extracts this same fetch logic to `useArchiveEvents`; apply the same generation guard there when integrating so the refactor does not reintroduce the race.

### DIFF
--- a/extension/website/archive/archive.tsx
+++ b/extension/website/archive/archive.tsx
@@ -44,7 +44,7 @@ function midnightWindowBounds(
   return { from: from.toISOString(), to: to.toISOString() };
 }
 
-const InternetMovement = () => {
+export const InternetMovement = () => {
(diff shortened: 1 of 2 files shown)


spencerc99:
## Code review

PR #303, opened 2026-07-12
Support primitive page data updates

spencerc99:
## Summary
- Support direct replacement and functional updates for primitive page-data roots.
- Read primitive functional updates from the live page-data store after remote Yjs replacements.
- Preserve object and array draft mutators, including terse returns such as d => d.count++.
- Document the contract and add working visit-counter examples to both starters.

## Root cause
Primitive values were held in the page-data map, but the existing replacement helper only mutated object and array proxies. Functional returns were ignored. The initial primitive implementation also retained a stale scalar in proxyByTagAndId after a peer replacement.

## Regression coverage
A two-document Yjs test seeds B with cached 0, applies A's remote update to 1, verifies B reads 1, then verifies B's value => value + 1 writes 2.

## PR #264 integration
#264 overlaps PageDataChannel.setData for server-gated writes. Keep this branch's root-replacement and PageDataSetter behavior, and derive its gated before snapshot from the same live primitive store value. Then mutate object drafts or use primitive updater returns before buildGatedWriteOps. The server already sends that value as a replace operation. I applied and type-checked that combined hunk against #264.

## Validation
- bun run build and bun run test in packages/playhtml: 364 tests passed
- bun run build and bun run test in packages/react: 44 tests passed
- Temporary #264 integration: packages/playhtml build and page-data suite passed

### DIFF
--- a/.changeset/page-data-primitive-roots.md
+++ b/.changeset/page-data-primitive-roots.md
@@ -1,0 +1,7 @@
+---
+"@playhtml/common": patch
+"playhtml": patch
+"@playhtml/react": patch
+---
+
+Page-data channels can now update primitive roots with functional updates that return the next value.
--- a/apps/docs/src/content/docs/data/page-data.mdx
+++ b/apps/docs/src/content/docs/data/page-data.mdx
@@ -37,10 +37,22 @@ function VisitCounter() {
   </TabItem>
 </Tabs>
 
+Channels can also hold a primitive value. Pass the next value directly or return
+the next value from a functional update:
+
+```js
+const viewCount = playhtml.createPageData("viewCount", 0);
+viewCount.setData((value) => value + 1);
+```
+
+For object and array channels, function updates are mutators: edit the draft in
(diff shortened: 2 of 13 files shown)


spencerc99:
## Code review

- **P2 [`packages/common/src/index.ts:399`](https://github.com/spencerc99/playhtml/blob/f56fc3f9e84bb241fbf1896e26f61e7bbb52bdf6/packages/common/src/index.ts#L399) makes nullable object channels lose contextual typing for functional updates.** Because the conditional type distributes over unions, `PageDataSetter<Item | null>` becomes a union containing incompatible callback signatures for `Item` and `null`. In strict TypeScript, a realistic update such as `channel.setData(value => value ?? { id: "first" })` now reports TS7006 because `value` is implicitly `any`; the previous setter signature contextually typed that parameter as the full `Item | null` union. Define the callback once over the full `T` (for example, a single `(value: T) => T | void` shape or overloads that preserve contextual typing) and add a declaration test for a nullable object channel.

PR #306, opened 2026-07-13
Fix element teardown and toggle readonly

spencerc99:
## Summary
- Clean up ElementHandler-installed listeners, including active document drag listeners, on destroy and allow a destroyed handler to reinitialize once.
- Forward CanToggleElement readOnly through the canonical dataSourceReadOnly prop.
- Keep CanPlayElement click and drag callbacks current after React rerenders without repeating setup.
- Add core and React regressions plus patch changesets for playhtml and @playhtml/react.

## Validation
- bun run test in packages/playhtml (366 tests)
- bun run test in packages/react (45 tests)
- bun run build-packages

## Docs and starters
- No updates: existing docs already describe CanToggleElement readOnly and its DOM attribute; starter templates do not demonstrate a read-only shared toggle.

### DIFF
--- a/.changeset/cleanup-element-listeners.md
+++ b/.changeset/cleanup-element-listeners.md
@@ -1,0 +1,5 @@
+---
+"playhtml": patch
+---
+
+Clean up built-in element listeners when an element is removed so remounting and dragging do not leave stale handlers active.
--- a/.changeset/cleanup-element-listeners.md
+++ b/.changeset/cleanup-element-listeners.md
@@ -2,4 +2,4 @@
 "playhtml": patch
 ---
 
-Clean up built-in element listeners when an element is removed so remounting and dragging do not leave stale handlers active.
+Clean up built-in element listeners when an element is removed so remounting and dragging do not leave stale handlers active, and install listeners when callbacks are added after setup.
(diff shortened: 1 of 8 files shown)


spencerc99:
## Code review

- **P1 [`packages/react/src/index.tsx:322`](https://github.com/spencerc99/playhtml/blob/9dda08b66887988a9f7b725785906cbfe8ede5bb/packages/react/src/index.tsx#L322-L344) cannot activate click or drag callbacks that were absent on the first render.** The rerender path only assigns `handler.onClick` / `handler.onDrag`, then `setupPlayElement(..., { ignoreIfAlreadySetup: true })` returns for the existing handler. But `ElementHandler` installs its DOM listeners only when the callback is present during reinitialization ([`elements.ts:238-299`](https://github.com/spencerc99/playhtml/blob/9dda08b66887988a9f7b725785906cbfe8ede5bb/packages/playhtml/src/elements.ts#L238-L299)). A component that initially renders without `onClick` or `onDrag` and adds it later therefore stores the new function but never receives the DOM event. Route callback updates through listener-aware handler logic, or install stable listeners independently of initial callback presence, and cover an absent-to-present rerender.

PR #314, opened 2026-07-16
Cycle archive playback through event batches

spencerc99:
## Summary

- cycle `/archive` through prefetched 1,000-event pages instead of replaying one fixed response forever
- fetch the next older matching page during playback and wrap to the newest page when the filtered history is exhausted
- move archive cycle completion out of `AnimatedTrails`, so batch handoff does not depend on a specific visualization
- share click-effect construction between live and archive trail renderers

## Why

The archive should keep moving through the matching history as each playback cycle finishes. Prefetching the next page keeps that handoff immediate, while generation checks prevent filter changes from applying stale responses.

## Relationship

This is the archive-playback follow-up to #313, which is now merged.

## Verification

- `bunx vitest run extension/website/shared/utils/__tests__/archiveEventBatches.test.ts extension/website/shared/utils/__tests__/trailAnimation.test.ts extension/website/shared/components/__tests__/LiveTrails.test.ts` (15 tests)
- `bunx tsc --noEmit` in `extension/website`
- `bun run build` in `extension/website`
- real `/archive/` browser run advanced through distinct older timestamp windows while rendering each new trail set

Visual evidence is attached in the PR conversation.

### DIFF
--- a/extension/website/archive/archive.tsx
+++ b/extension/website/archive/archive.tsx
@@ -65,11 +65,20 @@ const InternetMovement = () => {
     localStorage.setItem("movement_active_viz", JSON.stringify(activeVisualizations));
   }, [activeVisualizations]);
 
-  const { events, loading, error, dayCounts, refresh } = useArchiveEvents({
+  const {
+    events,
+    loading,
+    error,
+    dayCounts,
+    refresh,
+    advanceBatch,
+    batchKey,
+  } = useArchiveEvents({
     selectedDay,
     timeOfDay,
     serverDomain,
     activeVisualizations,
+    batchPlayback: true,
   });
 
   return (
@@ -105,6 +114,8 @@ const InternetMovement = () => {
         onSetFilters={setFilters}
         activeVisualizations={activeVisualizations}
         onSetActiveVisualizations={setActiveVisualizations}
+        playbackKey={batchKey}
+        onPlaybackCycleComplete={advanceBatch}
       />
     </>
   );
--- a/extension/website/shared/components/AnimatedClicks.tsx
+++ b/extension/website/shared/components/AnimatedClicks.tsx
@@ -19,6 +19,23 @@ export interface ScheduledClick {
   holdDuration?: number;
 }
 
+export const MAX_VISIBLE_CLICK_EFFECTS = 2000;
+
+export type VisibleClickEffect = ClickEffect & {
+  sourceId: string;
+};
+
+export function mergeClickEffects(
+  current: VisibleClickEffect[],
+  incoming: VisibleClickEffect[],
+): VisibleClickEffect[] {
+  const incomingSourceIds = new Set(incoming.map((effect) => effect.sourceId));
+  return [
+    ...current.filter((effect) => !incomingSourceIds.has(effect.sourceId)),
+    ...incoming,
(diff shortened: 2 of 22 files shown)


spencerc99:
## Preview evidence

**Archive playback advances through older 1,000-event batches while live remains a continuous WebSocket stream with trail-scoped click ripples.**

[Open preview](http://127.0.0.1:4177/archive/?animationSpeed=100000) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-mrnurz0b-232e7474) · Commit `bac9faf`

### Archive advances through older batches

[Watch the demo video](https://pr-evidence.spencerc99.workers.dev/media/ev-mrnurz0b-232e7474/archive-advances-through-older-batches/archive-advances-through-older-batches.mp4)

Accelerated playback visibly replaces each trail scene. Runtime diagnostics sampled ten distinct 1,000-event spans descending from 2:13 PM to 2:06 PM, confirming each handoff used the next older page rather than replaying the same data.

[Open captured page](http://127.0.0.1:4177/archive/?animationSpeed=100000) · 655×895

### Live stream and ripple lifecycle

[Watch the demo video](https://pr-evidence.spencerc99.workers.dev/media/ev-mrnurz0b-232e7474/live-stream-and-ripple-lifecycle/live-stream-and-ripple-lifecycle.mp4)

The live WebSocket remains connected while trails and ripples continue updating. After four seconds, 82 of 89 existing ripple positions remained while new ripples arrived, confirming ripples persist until their live trail is evicted.

[Open captured page](http://127.0.0.1:4177/portrait/) · 655×895

spencerc99:
## Code review

- **P1 [`extension/website/shared/hooks/usePlaybackCycle.ts:64`](https://github.com/spencerc99/playhtml/blob/bac9fafd5d61596c710c80917481556a17f5296a/extension/website/shared/hooks/usePlaybackCycle.ts#L64-L83) lets the batch clock resume from its pre-pause position even though the visual playback restarts from the beginning.** The hook merely stops incrementing `elapsedMs` while frozen, and `frozen` is absent from the effect dependencies, so unfreezing preserves the accumulated time. `AnimatedTrails` restarts its animation effect on every frozen-state change ([`AnimatedTrails.tsx:288`](https://github.com/spencerc99/playhtml/blob/bac9fafd5d61596c710c80917481556a17f5296a/extension/website/shared/components/AnimatedTrails.tsx#L288-L326)), while the other visualizers are unmounted behind `!paused`. Pausing 50 seconds into a 60-second batch and resuming therefore restarts the scene at zero but advances to the next batch about 10 seconds later. Restart the controller on unpause, or make the controller and visualizers share the same preserved clock.

- **P1 [`extension/website/shared/hooks/useArchiveEvents.ts:460`](https://github.com/spencerc99/playhtml/blob/bac9fafd5d61596c710c80917481556a17f5296a/extension/website/shared/hooks/useArchiveEvents.ts#L460-L477) cannot retry a failed prefetch after the request's built-in attempts are exhausted.** The rejection handler only clears `prefetchRequestRef`; mutating that ref does not render, and none of the effect dependencies change. At cycle completion `advanceBatch()` then returns false because `prefetched` is still null, also without changing state, so the prefetch effect never runs again and the same batch loops indefinitely until a manual refresh or filter change. Schedule a state-backed retry/backoff, or otherwise retrigger the prefetch effect after failure.

PR #316, opened 2026-07-22
quarantine-tape: interaction prototype (page + image taping)

spencerc99:
Interaction prototype for **quarantine tape** — a collective mechanic for cordoning off dead web (AI slop, SEO spam). This is the design/tuning artifact, not the shipped feature: a standalone test page with local state, no sync and no extension plumbing.

Try it: `bun run -C extension/website dev` → `/quarantine-tape-test/`

## What it does

**Page tape.** Equip a roll (cursor arms, the four viewport edges glow) → click an edge to anchor → pull taut tape to a second edge → commit. Strips go at any angle and always run fully off-screen. Strips accumulate per page: under 3 reads as provisional (faint, dashed), 3+ as cordoned.

**Image tape.** Hovering an image while armed outlines it; dragging across it lays an X of tape over just that element. The verdict is keyed by the image's `src`, so one mark paints every copy on the page — and it's the portable shape we'd need to mark the same artifact elsewhere later. Element tape re-measures each render, so it travels with the content on scroll (page tape stays viewport-fixed).

**Ripping.** Deliberately drag a stroke across a strip to slash it — a click is too easy for an act that contests collective judgment. Provisional tape falls in one slash; set tape takes three independent ones, with the requirement snapshotted at the first rip so teardown can't oscillate. Torn tape frays and stays rather than vanishing.

## Design notes

The unit of judgment is a **URL** — a page or an embedded resource — and tape is a *rendering* of that verdict. That split is what lets the same image be marked in more than one place later.

Two colorways, cosmetic for now: AI SLOP (amber/black) and SEO SPAM (white/red). Each strip is drawn horizontally in local space then rotated, so it looks identical at every angle; hazard slashes are explicit local geometry on the top/bottom edges rather than a `userSpaceOnUse` pattern, which tiles from the SVG origin and drifts across the text.

## Notes for review

- Nothing here is wired into the extension — `FLAGS.QUARANTINE_TAPE` stays false and the registry is untouched. Zero runtime impact.
- Images are natively draggable, which fires `dragstart` and never `mouseup`, silently swallowing the gesture; the native drag is suppressed while armed.
- Deferred: marking artifacts on remote surfaces (e.g. a slop image in search results). It needs a bulk "does this URL have a verdict?" lookup — we can't open a sync channel per thumbnail — so it's a separate phase.
- Paragraph/text taping is deferred too; text has no stable URL and needs selector anchoring.

Design doc: `internal-docs/specs/2026-06-15-quarantine-tape-design.md` (untracked, local).

PR #319, opened 2026-07-23
quarantine-tape: page tape (worker store + on-page feature)

spencerc99:
The collective mechanic for cordoning off dead web — string caution tape across a page to mark it as AI slop or SEO spam, reinforce by adding strips, rip by dragging across one. This is the **page-tape** cut, wired to a persistent worker store so a mark is there for the next visitor. Image taping is stacked on top in a follow-up PR.

## What's here

**Prototype** (`extension/website/quarantine-tape-test/`) — the tuned interaction + tape art, standalone test page.

**Worker store** (Supabase-backed):
- Migration: `quarantine_strips`, indexed by url, rips in a jsonb column.
- Three public routes: `GET /quarantine/verdict`, `POST /quarantine/strip`, `POST /quarantine/rip`. Rate-limited per IP, validated, url normalized server-side. 12 tests.

**Extension feature** (`features/social/quarantine-tape/`):
- Equip a tape roll from the satchel; string tape wall-to-wall; reinforce; rip.
- `quarantine-api.ts` is the storage seam — the only module that touches the backend, so the store can move later (playhtml rooms, a portable index) without touching the manager/rendering/gestures.
- Fetch-on-load, optimistic writes reconciled against the server. No live channel — the worker is the store.

## Design notes

The unit of judgment is a **URL**, and tape is a *rendering* of that verdict. Setness is derived from the strip count (3+ = cordoned). Ripping is a deliberate drag (a click's too easy); provisional tape falls in one slash, set tape takes three, with `rips_required` snapshotted on the first rip so teardown can't oscillate.

The url normalizer **keeps the query string** (drops only the hash) — an image src's identity often lives in the query, and collapsing distinct urls onto one key is irreversible. Tested.

## Notes for review
- `FLAGS.QUARANTINE_TAPE` ships **false** (inert for users; runs under the internal-dev override, e.g. the social-playground). Flip when ready to dogfood.
- Full worker suite green (60 tests). Extension type-checks clean.
- Rip races on the jsonb column are best-effort for a trial (documented in the route); a rips table with a unique constraint fixes it if it matters.
- Design doc + framework live in the Coda "Objects" page.

Supersedes #316 (which was prototype-only).


spencerc99:
## Code review

- **P1 [`supabase/migrations/20260723000516_extension_quarantine_tape.sql:29`](https://github.com/spencerc99/playhtml/blob/295b733c359dbeb29084daca1dabfdd85c4d884e/supabase/migrations/20260723000516_extension_quarantine_tape.sql#L29) leaves the new store directly writable through Supabase's public roles.** The migration grants `ALL` to `anon` and `authenticated` but never enables RLS for `quarantine_strips`. Anyone with the project's public Supabase credentials can therefore bypass the worker's validation and rate limiter and insert, update, or delete verdict rows directly. Enable RLS on the table and keep writes restricted to the service-role worker.

- **P1 [`extension/worker/src/routes/quarantine.ts:238`](https://github.com/spencerc99/playhtml/blob/295b733c359dbeb29084daca1dabfdd85c4d884e/extension/worker/src/routes/quarantine.ts#L238) does not bind the supplied URL to the strip being ripped.** The row is loaded and updated by `stripId` alone, while the first-rip threshold is counted using the caller-controlled `url` at line 262. A client can take a set strip ID returned by its real verdict, submit any valid URL with fewer than three rows, and permanently snapshot `rips_required = 1` on that strip. Filter the load/update by both ID and normalized URL, or derive the count from `strip.url` rather than the request body.

- **P2 [`extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts:303`](https://github.com/spencerc99/playhtml/blob/295b733c359dbeb29084daca1dabfdd85c4d884e/extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts#L303) discards a rip made while a strip POST is still in flight.** `ripStrip` explicitly returns for a temporary ID after mutating the optimistic strip, but reconciliation replaces that object with the server response, whose `rips` is still empty. On a slow request, disarming and slashing the freshly laid strip appears to work and then silently reverts. Carry pending rips onto the persisted ID and submit them after reconciliation, or keep the strip non-rippable until persistence completes.

- **P2 [`extension/src/features/social/registry.ts:9`](https://github.com/spencerc99/playhtml/blob/295b733c359dbeb29084daca1dabfdd85c4d884e/extension/src/features/social/registry.ts#L9) enables a user-facing extension feature without the required `extension/PENDING.md` release-note bullet.** The extension instructions require that note for changes under `extension/src/**`; add the user-facing quarantine-tape entry before merge.

spencerc99:
## Code review

- **P1 [`supabase/migrations/20260723000516_extension_quarantine_tape.sql:30`](https://github.com/spencerc99/playhtml/blob/2085ccbad772fcc9181b3f9fe43a0a40985d35ce/supabase/migrations/20260723000516_extension_quarantine_tape.sql#L30) leaves the new verdict table directly writable through the public Supabase roles.** The migration grants `ALL` to both `anon` and `authenticated` but never enables RLS, so anyone with the project's public Supabase credentials can bypass the worker's validation and rate limiter and insert, alter, or delete tape rows. Enable RLS explicitly and keep mutations available only to the service-role worker.

- **P1 [`extension/worker/src/routes/quarantine.ts:238`](https://github.com/spencerc99/playhtml/blob/2085ccbad772fcc9181b3f9fe43a0a40985d35ce/extension/worker/src/routes/quarantine.ts#L238) still lets the request URL determine another strip's teardown threshold.** The strip is loaded and updated by `stripId` alone, while line 262 counts the caller-controlled normalized URL. A client can submit a real strip ID with an unrelated URL that has fewer than three strips and permanently snapshot `rips_required = 1` on the target. Bind the lookup/update to both ID and URL, or derive the count from the loaded `strip.url`.

- **P2 [`extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts:303`](https://github.com/spencerc99/playhtml/blob/2085ccbad772fcc9181b3f9fe43a0a40985d35ce/extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts#L303) still drops a rip made while the strip POST is in flight.** `ripStrip()` mutates a temporary strip and returns at line 322, but reconciliation replaces that object with the server response, whose rip list is empty. On a slow request, ripping freshly laid tape appears to work and then silently reverts. Replay pending rips against the persisted ID after reconciliation, or keep temporary strips non-rippable until persistence completes.

PR #320, opened 2026-07-23
quarantine-tape: tape individual images (stacked on #319)

spencerc99:
**Stacked on #319** (page tape) — review/merge that first. This PR adds the ability to tape a single **image**, not just a whole page.

Marking a whole page is too blunt when the slop is one AI-generated image among real content. While a roll is armed, hovering an image outlines it; dragging across it lays an X of tape over that element.

## What's here

**Element store** (extends the worker):
- Migration: `quarantine_element_marks`, keyed by the image's **src** (query preserved), indexed by src.
- Three routes: **bulk** `GET /quarantine/element-verdict` (the on-page renderer asks about every tapeable image at once, capped at 100 — the pattern Phase-2 remote-surface rendering will reuse), `POST /quarantine/element-mark`, `POST /quarantine/element-rip`. 7 tests.
- Lifts `normalizeArtifactUrl` into `lib/artifactUrl` so page + element routes share one normalizer — the query-preserving behavior is migration-critical and must not diverge.

**Client** (extends the same feature):
- `ElementMark` type + three functions on the storage seam (still the only backend touchpoint).
- Hover-to-target, drag-across-image to tape, image ripping folded into the existing slash gesture.
- Element tape is **content-bound**: it re-measures the element's bounds each render, so it tracks the image through scroll/reflow (page tape stays viewport-fixed). Verdict keyed by src → one mark paints every copy of that image.

## Why this is a good property

Because the verdict is keyed by src, this is already the portable shape needed to mark the same image *elsewhere* (search-result thumbnails) in a later phase — no migration, just an added bulk-lookup surface, which this PR's `element-verdict` route already models.

## Notes for review
- The image gesture and the page-tape edge-commit gesture share the armed state but split on mousedown: a drag starting on a hovered image tapes the image; anywhere else the click-edge commit runs untouched.
- Host-page hover outline is a small `<style>` injected into `document.head` (the overlay's shadow-root CSS can't reach the page's own ``s), removed on destroy.
- Full worker suite green (67 tests). Extension type-checks clean.


spencerc99:
## Code review

- **P1 [`supabase/migrations/20260723002402_extension_quarantine_element_marks.sql:26`](https://github.com/spencerc99/playhtml/blob/a9c2dfac2f020189b57a4796c910d8d311b44300/supabase/migrations/20260723002402_extension_quarantine_element_marks.sql#L26) exposes unrestricted direct writes to the image-verdict table.** The migration grants `ALL` to `anon` and `authenticated` without enabling RLS, so a client with the project's public Supabase credentials can bypass every worker validation/rate-limit path and arbitrarily mutate or delete image marks. Enable RLS and leave mutations available only to the service-role worker.

- **P1 [`extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts:303`](https://github.com/spencerc99/playhtml/blob/a9c2dfac2f020189b57a4796c910d8d311b44300/extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts#L303) cannot discover the image underneath the armed overlay.** Arming sets the full-viewport shadow SVG to `pointer-events: auto` at line 152, so `document.elementFromPoint()` is retargeted to the high-z-index closed-shadow host/overlay rather than the host page's ``. `hoverTarget` therefore stays null and the image-drag branch never starts. Hit-test with `elementsFromPoint()` while skipping the injected host, or temporarily make the overlay non-hit-testable while resolving the underlying image.

- **P1 [`extension/worker/src/routes/quarantineElement.ts:183`](https://github.com/spencerc99/playhtml/blob/a9c2dfac2f020189b57a4796c910d8d311b44300/extension/worker/src/routes/quarantineElement.ts#L183) lets the request's `src` determine another mark's teardown threshold.** The mark is loaded/updated by `markId` alone, but line 204 counts the caller-controlled `src`. Supplying a real mark ID with an unrelated valid source that has fewer than three marks permanently snapshots `rips_required = 1`. Bind both queries to `markId + key`, or count using the loaded `mark.src`.

- **P2 [`extension/src/features/social/quarantine-tape/quarantine-api.ts:81`](https://github.com/spencerc99/playhtml/blob/a9c2dfac2f020189b57a4796c910d8d311b44300/extension/src/features/social/quarantine-tape/quarantine-api.ts#L81) sends every image source before the worker applies its 100-source cap.** On image-heavy pages, long CDN URLs can push this GET past browser/proxy URL limits, causing the entire verdict fetch to fail and render no existing tape. Apply the same cap client-side or request bounded batches.

- **P2 [`extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts:138`](https://github.com/spencerc99/playhtml/blob/a9c2dfac2f020189b57a4796c910d8d311b44300/extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts#L138) adds user-facing extension behavior without the required `extension/PENDING.md` bullet.** Add a release-note entry covering individual-image taping before merge.

spencerc99:
## Code review

- **P1 [`supabase/migrations/20260723002402_extension_quarantine_element_marks.sql:26`](https://github.com/spencerc99/playhtml/blob/fce625d0ff4f7447872c8e7f529ed0d75b7da36c/supabase/migrations/20260723002402_extension_quarantine_element_marks.sql#L26) leaves image verdicts directly writable through the public Supabase roles.** The migration grants `ALL` to `anon` and `authenticated` without enabling RLS, so a client with the public project credentials can bypass every worker validation/rate-limit path and arbitrarily mutate or delete image marks. Enable RLS explicitly and restrict mutations to the service-role worker.

- **P1 [`extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts:335`](https://github.com/spencerc99/playhtml/blob/fce625d0ff4f7447872c8e7f529ed0d75b7da36c/extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts#L335) cannot discover the image underneath the armed overlay.** Arming gives the full-viewport shadow SVG `pointer-events: auto` at line 41. A document-level `elementFromPoint()` therefore sees the shadow host/overlay hit rather than the host page's ``, leaving `hoverTarget` null and preventing the image-drag branch from starting. Resolve the underlying target with `elementsFromPoint()` while excluding the injected host, or temporarily make the overlay non-hit-testable during the lookup.

- **P1 [`extension/worker/src/routes/quarantineElement.ts:183`](https://github.com/spencerc99/playhtml/blob/fce625d0ff4f7447872c8e7f529ed0d75b7da36c/extension/worker/src/routes/quarantineElement.ts#L183) lets the request `src` determine another mark's teardown threshold.** The mark is loaded and updated by `markId` alone, while line 204 counts the caller-controlled normalized source. Supplying a real mark ID with an unrelated source that has fewer than three marks permanently snapshots `rips_required = 1`. Bind the lookup/update to both ID and source, or count using the loaded `mark.src`.

- **P2 [`extension/src/features/social/quarantine-tape/quarantine-api.ts:81`](https://github.com/spencerc99/playhtml/blob/fce625d0ff4f7447872c8e7f529ed0d75b7da36c/extension/src/features/social/quarantine-tape/quarantine-api.ts#L81) sends every image source before the worker applies its 100-source cap.** Image-heavy pages with long CDN URLs can exceed browser or proxy URL limits, causing the whole verdict request to fail and render no existing tape. Apply the cap client-side or request bounded batches.

- **P2 [`extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts:576`](https://github.com/spencerc99/playhtml/blob/fce625d0ff4f7447872c8e7f529ed0d75b7da36c/extension/src/features/social/quarantine-tape/QuarantineTapeManager.ts#L576) drops a rip made while an image-mark POST is in flight.** `ripElementMark()` mutates a temporary mark and returns at line 598, but commit reconciliation replaces that object with the server response, whose rip list is empty. The comment that the rip “rides along” is not implemented. Replay pending rips after the persisted ID arrives, or keep temporary marks non-rippable until then.

PR #321, opened 2026-07-23
Fix MovementCanvas audio crackling and live sound

spencerc99:
## Summary

- Crossfade oscillator changes so cursor instrument transitions do not crackle.
- Feed live cursor trails and click effects into the existing sound engine.
- Release live voices when playback pauses.
- Disconnect and delete per-trail audio state when a live trail is removed.

## Root cause

Archive playback changed oscillator waveforms immediately while voices were audible, which produced discontinuities in the signal. The live trail renderer did not send its draw frames or click effects to the sound engine. It also detached the engine while paused without releasing active voices and retained audio state after live trails expired.

## Verification

- `../../node_modules/.bin/vitest run --environment jsdom shared` (52 tests)
- `../../node_modules/.bin/tsc --noEmit`
- `bun run build`
- Verified sound enable, pause, and resume on the real `/portrait/` route with no console errors.

### DIFF
--- a/extension/website/shared/components/LiveTrails.tsx
+++ b/extension/website/shared/components/LiveTrails.tsx
@@ -162,6 +162,14 @@ export const LiveTrails: React.FC<LiveTrailsProps> = memo(
     const soundFramesRef = useRef<TrailSoundFrame[]>([]);
     const soundTrailIndicesRef = useRef<Map<string, number>>(new Map());
     const nextSoundTrailIndexRef = useRef(0);
+    const retiredSoundTrailIndicesRef = useRef<number[]>([]);
+    const queueSoundTrailRetirement = (trailId: string) => {
+      const soundTrailIndex = soundTrailIndicesRef.current.get(trailId);
+      if (soundTrailIndex !== undefined) {
+        retiredSoundTrailIndicesRef.current.push(soundTrailIndex);
+      }
+      soundTrailIndicesRef.current.delete(trailId);
+    };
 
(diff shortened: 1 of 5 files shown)


spencerc99:
## Code review

- **P1 [`extension/website/shared/components/MovementCanvas.tsx:1567`](https://github.com/spencerc99/playhtml/blob/0b353ecddadbe383b8cd632750f2b0732bddac88/extension/website/shared/components/MovementCanvas.tsx#L1567) detaches the live sound engine on pause without releasing its active voices.** `LiveTrails` returns from its frozen frame before calling `tick`, and switching the prop to `null` does not reset the existing Web Audio graph. Any sustained cursor voice therefore keeps sounding while the visualization is paused. Reset/release the engine before detaching it, or feed one empty frame when entering the paused state.

- **P2 [`extension/website/shared/components/LiveTrails.tsx:493`](https://github.com/spencerc99/playhtml/blob/0b353ecddadbe383b8cd632750f2b0732bddac88/extension/website/shared/components/LiveTrails.tsx#L493) gives every live trail a monotonically increasing sound index without retiring the corresponding engine state.** Removing a trail only deletes the local ID mapping, while `SoundEngine.releaseVoice()` nulls oscillators but leaves the `voices` entry, `prevPositions`, and connected gain/filter/panner nodes retained. A long-running `/portrait/` stream therefore accumulates state for every historical trail. Add an explicit trail-retirement path that disconnects and deletes all per-index engine state, or safely reuse released indices.

PR #324, opened 2026-07-23
Fix extension event storage integrity and aggregate growth

spencerc99:
## Summary

- Retry content-script event storage with capped backoff when background storage reports a failed write or the runtime message rejects.
- Replace growing URL and navigation ID arrays in IndexedDB aggregates with exact URL membership records and bounded counts.
- Migrate existing aggregates in place so retained history survives the storage upgrade.
- Add regression coverage and extension release notes for both fixes.

## Root cause

`EventBuffer` treated any fulfilled `STORE_EVENTS` message as success, including `{ success: false }`. It removed those events from memory and then started an upload flush even though IndexedDB had rejected the write.

`LocalEventStore` serialized every URL and processed navigation ID into each aggregate. The global aggregate grew for the lifetime of the extension and was copied on every navigation, making storage work increase with browsing history.

## Impact

Failed local writes no longer silently discard newly collected activity. Aggregate updates now stay bounded as browsing history grows while preserving exact unique-page counts and previously retained statistics.

## Verification

- Extension Vitest suite: 65 files, 408 tests passed.
- Targeted TypeScript checks for `EventBuffer`, `LocalEventStore`, and their regression tests passed.
- `git diff --check` passed.

Screenshots are not applicable because this PR does not change a visual surface.


spencerc99:
## Code review

- **P1 [`extension/src/storage/EventBuffer.ts:111`](https://github.com/spencerc99/playhtml/blob/db65be3a7f536ec418a2c9887ac15df1bcd7ef55/extension/src/storage/EventBuffer.ts#L111-L113) leaves failed batches queued without scheduling any retry.** After a `{ success: false }` response, these lines restore the events but call neither `scheduleStoreFlush()` nor `scheduleBatch()`; the storage timer was cleared and the batch timer has already fired. If no later event arrives or caller manually invokes `flushBatch()`, the batch stays only in memory and is lost on navigation or tab close—the same data-loss path this PR is meant to fix. Schedule a bounded/backoff storage retry whenever a batch is requeued (including rejected messages), and test it by advancing the retry timer instead of manually calling `flushBatch()` a second time.

PR #325, opened 2026-07-24
extension: button, svg-icon, and cursor scrap kinds

spencerc99:
## Summary

Stacked on #323. Extends internet scraps with three new object kinds — all riding the existing `element` event type's payload union, so **zero changes to `@playhtml/extension-types`** (no changeset, no worker bump; this is the payoff of the kind-discriminant design).

- **`button`** — semantic candidates only (`<button>`, `input[type=submit|button]`, `[role=button]`), same visibility gate as images, per-page cap 20. Captured by **reconstruction** (label text + allowlisted computed-style subset + optional sanitized inner svg), never raw outerHTML — captured markup renders in a privileged extension page, so arbitrary HTML replay is off the table. Gradient backgrounds kept, `url(...)` backgrounds never.
- **`svg-icon`** — inline `<svg>` (excluding those inside button candidates), 12–400px, per-page cap 20. Serialized with in-document `<use>` refs inlined (external/unresolvable → skip), `currentColor` baked to computed values, run through an allowlist sanitizer (elements whitelist, external-resource rejection incl. `url()` in any attribute, `on*`/`javascript:` stripping), 20KB cap.
- **`cursor`** — passive 500ms-throttled mouseover sampling of computed `cursor: url(...)` + hotspot. `data:` URIs allowed (self-contained cursors never rot), `blob:` skipped, uncapped — rare finds.

`GET_SCRAPS` now returns a discriminated union with a per-kind `key` (src / style-hash / markup-hash / url) used for both capture-time and render-time dedup. The collage renders per-kind tiles: buttons as styled `<span>`s (native button UA styling can't fight the captured look), icons via sanitized markup, cursors as tiny pixelated tiles; provenance cards gain a kind label.

Consumption affordances (equipping cursors via the satchel, placing scraps) are deliberately NOT in this PR — affordances derive from `kind` at consumption time per the plan.

## Verification

- 424/424 extension tests (new coverage: button text/size filters + input labels, svg sanitizer strips + `use` resolution + byte cap, cursor parsing/throttle/dedup, union mapping + keys)
- Typecheck: exactly the 40 pre-existing baseline errors, none referencing changed files; website tsconfig clean
- Chrome production build succeeds
- Sanitizer hand-reviewed: allowlist-based, resolve-then-sanitize ordering, cycle-guarded `use` resolution, external-URL rejection across attribute values

## Notes

- No `PENDING.md` bullet (ships dark behind the same `FLAGS.SCRAPS` / dev-toggle gate — per the release-notes rule).
- No docs/starter changes needed: no public core-package API touched.
- Retarget this PR to `main` after #323 merges.

### DIFF
--- a/extension/src/__tests__/background-scraps.test.ts
+++ b/extension/src/__tests__/background-scraps.test.ts
@@ -1,10 +1,34 @@
 // ABOUTME: Verifies the background scrap query response used by extension rendering surfaces.
-// ABOUTME: Guards newest-first query limits and stored-event metadata mapping.
+// ABOUTME: Guards union mapping, stable render keys, query limits, ordering, and unknown kinds.
 
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { CollectionEvent } from "@playhtml/extension-types";
+import { hashScrapString, serializeScrapStyles } from "../collectors/scrapUtils";
 
 const originalDefineBackground = (globalThis as any).defineBackground;
+
+function createEvent(
+  id: string,
+  ts: number,
+  data: Record<string, unknown>,
+  domain: string | undefined = "example.com",
+): CollectionEvent {
+  return {
+    id,
+    type: "element",
+    ts,
+    data,
+    meta: {
+      pid: "pid",
+      sid: "sid",
+      url: `https://example.com/${id}`,
+      vw: 1024,
+      vh: 768,
+      tz: "America/Los_Angeles",
+    },
+    domain,
+  } as CollectionEvent;
+}
 
 describe("background scrap queries", () => {
   beforeEach(() => {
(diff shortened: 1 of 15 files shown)


spencerc99:
## Code review

- **P1 [`extension/src/collectors/ScrapCollector.ts:393`](https://github.com/spencerc99/playhtml/blob/766cf175cea250db56584f2cacf9fa92acdac30c/extension/src/collectors/ScrapCollector.ts#L393) lets a mixed gradient/URL background bypass the URL exclusion.** `GRADIENT_PATTERN` only verifies that the computed value starts with a gradient, so a valid multilayer value such as `linear-gradient(...), url("https://example.com/tracker.png")` passes and the entire value is stored. The renderer then spreads that value back onto the extension page at [`ScrapCollage.tsx:502`](https://github.com/spencerc99/playhtml/blob/766cf175cea250db56584f2cacf9fa92acdac30c/extension/website/shared/components/ScrapCollage.tsx#L502), causing the page-controlled URL to be fetched/rendered despite the stated allowlist. Reject any value containing `url(`, or parse every layer and retain only gradients.

- **P1 [`extension/src/collectors/ScrapCollector.ts:458`](https://github.com/spencerc99/playhtml/blob/766cf175cea250db56584f2cacf9fa92acdac30c/extension/src/collectors/ScrapCollector.ts#L458) archives page-controlled cursor URLs without either a size limit or a per-page count limit.** Every 500 ms a page can expose a unique `data:` cursor; lines 464-487 accept the full string and emit it into the persistent event pipeline. A hostile page can therefore make hovered elements carry unique multi-megabyte data URLs and rapidly exhaust the extension IndexedDB/upload path. Enforce a strict encoded-byte limit and a per-page cursor cap before calling `emit()`.

spencerc99:
Added the exploration surfaces for testing (all behind nothing — plain website pages, deploy with wewere.online):

- `/scraps-preview/` — collage with synthetic data for all four kinds + the new kind-filter chips (also wired into the extension's scraps.html page)
- `/scraps-grid/` — variable-footprint cabinet inventory prototype (drag to rearrange, r rotates while dragging)
- `/scraps-pile/` — physics junk-crate prototype (drag/throw; tuning constants at the top of the file)

Known issue on the pile: in some settle configurations one or two items keep slowly rotating in place (~5-10 deg/s) instead of fully sleeping — most bodies rest perfectly still. Root causes fixed so far: sleep-mass inconsistency in the impulse solver, floor rolling coupling, degenerate 0x0 ResizeObserver measurements collapsing spawn positions. The residual slow-spin is intermittent and cosmetic; punting further tuning until the design direction is picked.

PR #334, opened 2026-07-25
playhtml: make register the vanilla custom element API

### LINKED ISSUE #254: Consolidate view-API plumbing with existing element setup (post-#190 follow-up)
Follow-up from review of #190 (reactive `view` API). These are simplification/consolidation opportunities deferred out of #190 to keep that PR's runtime diff stable for merge (per the "leave these simplification changes until we confirm this is good to merge" note on the PR). The view API itself ships in #190; this issue tracks the cleanup.

### 1. Collapse `register()` / `setupPlayElement` into one path
`registerPlayElement(id, init)` stamps the initializer onto the element, ensures `can-play`, and calls `setupPlayElementForTag` — it already converges with the normal setup path. Explore making them one function ("if it's the view API handle the view API, otherwise the imperative path") so there's a single entry point.

### 2. Generalize the `register(id, …)` ergonomics to imperative can-play
The id-based `playhtml.register(id, init)` entry point (no need to put `can-play` on the element, and you can't usefully add the attribute alone without also defining `defaultData`/handlers) is arguably nicer than the set-props-on-the-element pattern. Consider offering this for imperative can-play too, not just the view API.

### 3. Reconsider `stampRegistrationOntoElement` (`Object.assign(element, init)`)
Setting initializer fields directly on the DOM element is simple but implicit — "you don't know what's happening." Consider passing `init` to the handler directly instead of round-tripping through element props. (Note this is the same mechanism React's `CanPlayElement` uses via `getCustomElementProps`, so any change should keep that path working.)

### 4. Share `definePlayCapability` with the existing new-tag registration
`define(name, init)` writes to the same `capabilitiesToInitializer` registry that `init({ extraCapabilities })` uses, then upgrades existing elements. Factor the validate → register → upgrade-existing logic into one shared helper used by both.

### 5. Unify the view-descendant MutationObserver with can-mirror's observer
`setupViewDescendants` (bind/teardown capability descendants emitted by a view) and the `can-mirror` DOM-sync observer both watch `{ childList, subtree }` and do similar reconciliation. See whether a shared observer/utility can back both, despite their different purposes (view = bind/teardown capability handlers; mirror = sync DOM ↔ CRDT).

### Notes
- Items 1–4 are pure refactors of code introduced in #190 with real regression surface — they need their own tests + review.
- Item 5 touches both the new code and existing can-mirror, so it's the highest-risk and should be scoped carefully.

spencerc99:
## Summary

- make `playhtml.register(elementOrId, initializer)` the recommended vanilla API for one-off custom elements using either `updateElement` or `view`
- accept an existing HTML element to preserve the `setupPlayElement(element)` call shape, or accept an id before the element exists
- route `register()`, `define()`, and configured capabilities through shared validation and setup paths without copying registered initializers onto DOM elements
- migrate public docs, runnable recipes, website examples, both starter templates, and the bundled element-building guide to register-first usage
- reuse the same mutation observer helper for registered view descendants and `can-mirror`

## Soft deprecation

Direct DOM-property setup remains supported for compatibility, with no runtime warning or removal in this PR. New vanilla code should move these fields into the initializer passed to `register`: `defaultData`, `defaultLocalData`, `myDefaultAwareness`, `updateElement`, `view`, `updateElementAwareness`, `onClick`, `onDrag`, `onDragStart`, `onMount`, `resetShortcut`, `debounceMs`, and `isValidElementForTag`.

`register(element, initializer)` requires an HTML element with a stable, non-empty id and binds that node directly. `register(id, initializer)` keeps the existing deferred form and binds when an element with that id appears.

`setupPlayElement()` is not deprecated. It remains the setup path for dynamically inserted built-in or `define()` capabilities. Registered custom elements do not need a separate setup call.

React's `withSharedState` API remains supported, including its internal property-based binding. `register()`, `define()`, handles, and imperative `updateElement` are documented as supported APIs; the declarative `view` renderer and lit-html helpers remain experimental.

## Why

Issue #254 started as duplicated registration and observer plumbing, but the public direction is also important: vanilla custom-element behavior should live in one initializer object rather than being spread across properties assigned to a DOM node. The element overload keeps the familiar element-first setup shape without reviving property assignment. Keeping the property path as a soft compatibility layer lets existing sites and React continue working while docs and examples consistently teach `register()`.

## Validation

- `bun build-packages`
- `bun run --cwd packages/playhtml test` — 507 tests
- `bun run --cwd packages/react test` — 50 unit tests and 1 integration test
- full docs test suite — 41 tests
- `bun build-site`
- `bunx tsc` in `website`
- targeted story renderer test
- `bunx vite build` in both starter templates
- live browser verification: `register(element, initializer)` on the homepage candle, overload docs, HTML and React starter reaction flows, one-word story presence, and cinderblock live sync

Visual evidence: https://github.com/spencerc99/playhtml/pull/334#issuecomment-5097911667

Closes #254

### DIFF
--- a/.changeset/calm-views-register.md
+++ b/.changeset/calm-views-register.md
@@ -3,4 +3,4 @@
 "@playhtml/common": patch
 ---
 
-Route programmatic element and capability registration through the standard setup paths. Registered initializers now bind without being copied onto DOM elements, while React's property-based `can-play` setup remains supported.
+Make `playhtml.register(id, initializer)` the recommended vanilla API for custom elements using either `updateElement` or `view`. Registered initializers now bind through the standard setup path without being copied onto DOM elements, while direct element-property configuration remains supported for compatibility.
--- a/.changeset/calm-views-register.md
+++ b/.changeset/calm-views-register.md
@@ -1,0 +1,6 @@
+---
+"playhtml": patch
+"@playhtml/common": patch
+---
+
+Route programmatic element and capability registration through the standard setup paths. Registered initializers now bind without being copied onto DOM elements, while React's property-based `can-play` setup remains supported.
--- a/.changeset/calm-views-register.md
+++ b/.changeset/calm-views-register.md
@@ -3,4 +3,4 @@
 "@playhtml/common": patch
 ---
 
-Make `playhtml.register(id, initializer)` the recommended vanilla API for custom elements using either `updateElement` or `view`. Registered initializers now bind through the standard setup path without being copied onto DOM elements, while direct element-property configuration remains supported for compatibility.
+Make `playhtml.register(elementOrId, initializer)` the recommended vanilla API for custom elements using either `updateElement` or `view`. Callers can pass an existing HTML element or register by id before the element exists. Registered initializers bind through the standard setup path without being copied onto DOM elements, while direct element-property configuration remains supported for compatibility.
--- a/apps/docs/src/components/playground/recipes/__tests__/basic-recipes.test.ts
+++ b/apps/docs/src/components/playground/recipes/__tests__/basic-recipes.test.ts
@@ -67,28 +67,44 @@ describe("basic canonical recipes", () => {
     );
   });
 
-  it("configures can-play elements before initialization", () => {
+  it("registers can-play elements before initialization", () => {
     for (const recipe of [sharedCounterRecipe, sharedGuestbookRecipe]) {
+      const registerIndex = recipe.html.indexOf("playhtml.register(");
       const initIndex = recipe.html.indexOf("await playhtml.init");
-      expect(recipe.html.indexOf(".defaultData")).toBeLessThan(initIndex);
-      expect(recipe.html.indexOf(".updateElement")).toBeLessThan(initIndex);
+      expect(registerIndex).toBeGreaterThan(-1);
+      expect(registerIndex).toBeLessThan(initIndex);
+      expect(recipe.html.indexOf("defaultData:", registerIndex)).toBeLessThan(
+        initIndex,
+      );
+      expect(recipe.html.indexOf("updateElement:", registerIndex)).toBeLessThan(
+        initIndex,
+      );
     }
   });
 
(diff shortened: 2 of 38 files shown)


spencerc99:
## Preview evidence

**playhtml.register now accepts either an existing HTML element or its id while preserving the register-first setup path.**

[Open preview](http://127.0.0.1:4321/docs/reference/view-api/) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-ms3ve9v8-110d4e23) · Commit `e6f37a8`

### Element or id registration

Element or id registration

The registration reference documents the element overload, its id requirement, and the deferred id form side by side.

[Open captured page](http://127.0.0.1:4321/docs/reference/view-api/#playhtmlregisterelementorid-init) · 1280×720

### Vanilla starter interaction

Vanilla starter interaction

The HTML starter's registered reaction view renders and updates the shared count from 0 to 1.

[Open captured page](http://127.0.0.1:4174/) · 1280×720

### React starter remains supported

React starter remains supported

The withSharedState starter path still renders and updates the reaction count after the vanilla API migration.

[Open captured page](http://127.0.0.1:4175/) · 1280×720

spencerc99:
## Code review

- **P1 [`packages/playhtml/src/index.ts:3055`](https://github.com/spencerc99/playhtml/blob/e6f37a839a1a246cca4f8f10b86e3279fd8034ef/packages/playhtml/src/index.ts#L3055) leaves an id registration made after initialization unbound when its element is inserted later.** In the string form, when `document.getElementById()` returns null, this stores the initializer and skips `setupPlayElement`; there is no DOM-insertion observer elsewhere that revisits registered ids. The changed docs now promise that `register` binds the element “when it appears” and explicitly say not to call `setupPlayElement`, so `await playhtml.init(); playhtml.register("late", init); document.body.append(divWithThatId)` never creates a handler or runs the renderer until some caller manually re-walks/setup the node. Add an insertion path for registered ids (and cover the after-init/deferred-element case), or keep the manual setup requirement in the API contract.

PR #337, opened 2026-07-27
extension: fix leaked cleanups in content script (custom-site socket, emotes, global features)

spencerc99:
## Summary

Fixes a pre-existing resource leak in `extension/src/entrypoints/content.ts`: `initCustomSite()` returns a cleanup function that `destroy()`s the Wikipedia lobby's `createPresenceRoom` socket and its listeners, but the return value was discarded — so the socket leaked on every page navigation. Two more cleanups in the same file (`emoteCleanup`, `globalCleanup`) were stored but never invoked, the identical dropped-cleanup pattern, so they're wired in too.

## Changes

- Store the cleanup returned by `initCustomSite()` in a new `customSiteCleanup` field.
- Add an idempotent `teardown()` method that runs and nulls `customSiteCleanup`, `emoteCleanup`, and `globalCleanup`.
- Register teardown on `pagehide` (covers bfcache/normal unload), `beforeunload` (fallback), and `ctx.onInvalidated` (extension reload/update while the page stays open). `ctx` is now threaded into `main(ctx)`; the `ctx?.` optional call keeps the existing tests (which invoke `main()` bare) passing while WXT always supplies it in production.

## Verification

- `bun build-packages` then extension suite: **67 files, 428 tests, all passing**.
- `tsc`: no errors in the changed file (pre-existing `worker/` and `wxt.config.ts` errors exist on `main`, untouched).

## Notes

- No changeset: change is under `extension/`, not `packages/`.
- No `PENDING.md` bullet: internal resource-leak fix with no user-visible effect, per the release-note criteria in `extension/CLAUDE.md`.
- No screenshots: no user-facing surface changes — this only releases sockets/listeners that were previously orphaned on unload.

### DIFF
--- a/extension/src/__tests__/content-lifecycle.test.ts
+++ b/extension/src/__tests__/content-lifecycle.test.ts
@@ -1,0 +1,172 @@
+// ABOUTME: Regression tests for content-script page-lifecycle teardown/restore wiring.
+// ABOUTME: Verifies bfcache pageshow restores collaboration and listeners are not one-shot.
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const storageGet = vi.hoisted(() => vi.fn());
+const storageSet = vi.hoisted(() => vi.fn());
+const runtimeSendMessage = vi.hoisted(() => vi.fn());
+
+vi.mock("webextension-polyfill", () => ({
+  default: {
+    storage: {
+      local: {
+        get: storageGet,
+        set: storageSet,
+      },
+      onChanged: {
+        addListener: vi.fn(),
+      },
+    },
+    runtime: {
+      getURL: vi.fn((path: string) => `chrome-extension://test/${path}`),
+      sendMessage: runtimeSendMessage,
+      onMessage: {
+        addListener: vi.fn(),
+      },
+    },
+  },
+}));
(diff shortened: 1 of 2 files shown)


spencerc99:
## Code review

- **P1 [`extension/src/entrypoints/content.ts:1361`](https://github.com/spencerc99/playhtml/blob/fc900c86682b74387a3dc9d8adb5ba870edd2288/extension/src/entrypoints/content.ts#L1361) permanently tears down collaborative features when a page enters the back/forward cache.** `pagehide` also fires when a page is cached for back/forward navigation. Because this handler is `{ once: true }`, it clears the custom-site, emote, and global resources and removes itself, but there is no `pageshow` path to recreate them. Returning to a cached Wikipedia page therefore resumes the same initialized content script with its collaboration resources gone and no teardown listener for the next navigation. [Chrome lifecycle guidance](https://developer.chrome.com/docs/web-platform/deprecating-unload#recommended-alternatives) specifically notes that resources cleaned up on `pagehide` may need restoration when the page returns. Handle `PageTransitionEvent.persisted` and reinitialize on `pageshow`, or otherwise preserve restorable state while retaining normal unload and invalidation cleanup.

PR #339, opened 2026-07-29
partykit: room load/compaction circuit breaker with manual-first quarantine

spencerc99:
## Incident context

Three rooms with oversized persisted docs (7-13MB base64) put their Durable Objects into a permanent OOM crash loop: the empty-room compaction alarm loads the doc, `buildCompactedDocument` holds 3-5 representations at once, the isolate OOMs, and Cloudflare auto-retries the failed alarm within seconds. Because the memory limit is per-isolate (shared by many co-located rooms), each reset also killed unrelated rooms' connections — including the playhtml.fun homepage's presence socket, which is how it surfaced. Two of the three rooms were fixed by external compaction (13MB→152KB and 7.8MB→488KB, lossless-verified); the remaining two hold mostly real content and still crash-loop today. This PR makes the server structurally immune to the loop.

## Policy (as decided in review)

| Trigger | Response | Room state |
|---|---|---|
| Doc > 10MB at load | Warn only, hydrate anyway | Normal |
| Doc > 4MB at compaction | Skip in-DO compaction, park schedule, log external runbook once (durable once-guard) | **Normal** — loads, serves, persists |
| Risky work fails (load/alarm) | Durable counters + exponential backoff (1m → 5m → 20m → 1h → 6h → 24h cap); deferred rooms 503 + Retry-After, zero Supabase reads, single-flight in-place retry at deadline | Deferred, self-heals |
| 8 consecutive vanished-mid-work failures | Auto-quarantine (last resort) | Transient: serves visitors ephemeral-only, all persistence writes blocked, alarms parked |
| `POST admin/quarantine-set` / `-clear` (+ Cloudflare KV flag, checked BEFORE hydration) | Manual quarantine — the primary path | Transient |
| `POST admin/restore-raw-document` | Replaces doc; auto-clears park/quarantine only if the new doc is under the ceiling | Normal |

Size alone never blocks a load or quarantines a room (four healthy production rooms sit in the 6-8MB band). Quarantine never touches the persisted document — data-safety is enforced at every Supabase write function (`assertNotQuarantined`), enumerated and verified in review.

## Key design points

- **The gate runs in `onStart` BEFORE `super.onStart()`** — partyserver's `alarm()`/`fetch`/websocket paths all hydrate via `ensureInitialized → onStart → onLoad` before any handler runs, so this is the only placement that intercepts every entry path. Caught in fresh review (guards originally in `onAlarm` were unreachable for hydration-OOM rooms); regression-proofed by a new test class that drives the real platform entry points (`server.alarm()`, `server.fetch()`) instead of calling hooks directly.
- **Manual quarantine works on rooms that can't start**: the flag lives in Cloudflare KV (namespaces created for prod/preview/staging, ids wired in wrangler.jsonc) and is read pre-hydration. KV reads fail open (operator tool, not a correctness gate); presence of the key quarantines even with an empty value (fail-safe — release by deleting the key, not writing empty). Emergency stop for an unreachable room: `bunx wrangler kv key put --binding QUARANTINE_CONTROL "quarantine:<room-id>" "reason" --remote`.
- **Observed exceptions are not OOM evidence**: the failure counters only accumulate on true vanished-mid-work attempts, so a Supabase outage logs loudly at normal cadence instead of slowly quarantining healthy rooms.
- Backoff deadlines/counters are durably written before risky work, which is what absorbs Cloudflare's seconds-apart alarm retries.

## Review process

Implemented against a live incident, then hardened through: Spencer's line-level review (4 required changes incl. the 503 contract and KV control plane), a fresh DO-lifecycle review (caught the F1 entry-path blocker + 7 more), and a silent-failure audit (caught the eternal-503 trap, the un-awaited admin router swallowing errors, and 8 more). Every finding fixed or explicitly documented in code; each mechanism mutation-tested.

## Verification

- 198 tests across 17 partykit files, run from the branch worktree; `tsc -p partykit` clean
- Data-safety: persisted-row byte-identity asserted across autosave, compaction, hard reset, restore, and force-save under quarantine
- Platform-entry tests: alarm on a deferred room performs zero Supabase reads; KV-flagged room never hydrates; 5 concurrent requests at a deferral deadline produce exactly 1 read

No packages/ changes — server-only, no changeset. Deploy: staging first (`bun deploy-server:staging`), verify, then `bun deploy-server`. On production deploy the two remaining crash-looping rooms stop looping immediately (compaction-skip parks their alarms); they then need one external compaction each (runbook in the parked-room log line).

### DIFF
--- a/partykit/__tests__/quarantineControl.test.ts
+++ b/partykit/__tests__/quarantineControl.test.ts
@@ -1,0 +1,42 @@
+// ABOUTME: Verifies pagination and filtering for the global quarantine index.
+// ABOUTME: Ensures the admin list reports every current KV-backed room once.
+import { describe, expect, test } from "bun:test";
+import { listQuarantinedRooms } from "../quarantineControl";
+
+describe("listQuarantinedRooms", () => {
+  test("reads every page and ignores keys deleted during the listing", async () => {
+    const values = new Map([
+      ["quarantine:room-b", "automatic alarm failures"],
+      ["quarantine:room-a", "operator stop"],
+    ]);
+    const kv = {
+      async list({ cursor }: { cursor?: string }) {
+        if (!cursor) {
+          return {
+            keys: [
+              { name: "quarantine:room-b" },
+              { name: "quarantine:deleted-room" },
+            ],
+            list_complete: false as const,
+            cursor: "next",
+            cacheStatus: null,
+          };
+        }
+
+        return {
+          keys: [{ name: "quarantine:room-a" }],
+          list_complete: true as const,
+          cacheStatus: null,
+        };
+      },
+      async get(key: string) {
+        return values.get(key) ?? null;
+      },
+    } as unknown as KVNamespace;
+
+    await expect(listQuarantinedRooms(kv)).resolves.toEqual([
+      { roomId: "room-a", detail: "operator stop" },
+      { roomId: "room-b", detail: "automatic alarm failures" },
+    ]);
+  });
+});
--- a/partykit/__tests__/quarantineLoad.test.ts
+++ b/partykit/__tests__/quarantineLoad.test.ts
@@ -151,6 +151,21 @@ function createRoom(name = "example-room") {
   return { room: buildRoom(storage, name), storage };
 }
 
+/**
+ * Mirrors the platform's start sequence without the full Server base class:
+ * the pre-hydration gate in onStart, then hydration only if it allows it.
+ * Tests that need the genuine entry points live in quarantinePlatformEntry.
+ */
+async function startRoom(room: any): Promise<void> {
+  await room.applyExternalQuarantineFlag();
+  if (room.isQuarantined()) {
+    await room.enterQuarantineRuntimeState();
+    return;
+  }
+  if (await room.shouldDeferLoad()) return;
+  await room.onLoad();
+}
+
 // Models a Durable Object restart: a fresh instance over the same storage.
 function restartRoom(storage: FakeStorage, name = "example-room") {
   return buildRoom(storage, name);
@@ -204,7 +219,7 @@ describe("load path", () => {
     persistedRow.document = COMPACT_LETHAL_DOCUMENT;
     const { room } = createRoom();
 
-    await room.onLoad();
+    await startRoom(room);
 
     expect(room.isQuarantined()).toBe(false);
     expect(room.isPersistenceAvailable()).toBe(true);
@@ -215,7 +230,7 @@ describe("load path", () => {
     persistedRow.document = SMALL_DOCUMENT;
     const { room, storage } = createRoom();
 
-    await room.onLoad();
+    await startRoom(room);
 
     expect(room.isQuarantined()).toBe(false);
     expect(room.document.getMap("play").get("greeting")).toBe("hello");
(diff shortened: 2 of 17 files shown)


spencerc99:
## Code review

- **P1 [`partykit/party.ts:251`](https://github.com/spencerc99/playhtml/blob/d511b257e8c628a93500a37df2ec4719bce4094a/partykit/party.ts#L251-L253) skips the base Yjs startup for the cold quarantine and recovered-load paths.** A KV-flagged start returns before `super.onStart()`, but that base method is what attaches document-update broadcasts, awareness broadcasts, and the debounced `onSave` listener. The advertised transient room therefore does not actually sync changes between visitors. The same gap remains after a load deferral: [`attemptDeferredReload()`](https://github.com/spencerc99/playhtml/blob/d511b257e8c628a93500a37df2ec4719bce4094a/partykit/party.ts#L964-L974) calls only `onLoad()`, so a recovered normal room accepts connections without those base observers and subsequent edits are neither broadcast nor persisted. Let quarantined starts pass through the safe `onLoad` early return while still completing base startup, and make deferred recovery complete the same base/bridge initialization before declaring the room recovered.

- **P1 [`partykit/party.ts:2689`](https://github.com/spencerc99/playhtml/blob/d511b257e8c628a93500a37df2ec4719bce4094a/partykit/party.ts#L2684-L2693) can hydrate the room before an emergency control route runs.** `getLoadDeferredResponse()` is called unconditionally before `isLoadControlRoute` is applied; once the deadline has passed, that method starts `attemptDeferredReload()`. An operator calling `quarantine-set` or `quarantine-status` at that point can therefore trigger the risky hydration/OOM and never reach the handler, contradicting the guarantee that these routes stay reachable throughout backoff. Bypass deferred-load recovery entirely for the three control routes, and only call `getLoadDeferredResponse()` for other admin requests.

- **P1 [`partykit/party.ts:1084`](https://github.com/spencerc99/playhtml/blob/d511b257e8c628a93500a37df2ec4719bce4094a/partykit/party.ts#L1080-L1092) exposes an empty live document after clearing a cold quarantine or load deferral.** `clearQuarantine()` removes the quarantine, counters, deadline, and in-memory deferral immediately, but it neither hydrates the persisted snapshot nor causes a new initialization. On the exact broken-room path this feature targets, the current Y.Doc was never loaded, so the next request or WebSocket is admitted against an empty room even though the real data remains in Supabase. Keep the load guard active until a full guarded hydration/startup succeeds, or complete that initialization as part of clearing before allowing normal traffic.

spencerc99:
## Preview evidence

**The admin dashboard now distinguishes delayed automatic compaction retries from per-room compaction disablement after three vanished attempts.**

[Open preview](http://127.0.0.1:5173/admin-compaction-visual.html) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-ms6kwv3d-3eca5a3a) · Commit `c6103be`

### Automatic compaction recovery states

Automatic compaction recovery states

The real admin panel shows a delayed retry after one vanished attempt and disables only automatic compaction after the third, while the room load and alarm ledgers remain healthy.

[Open captured page](http://127.0.0.1:5173/admin-compaction-visual.html) · 1440×1368

PR #345, opened 2026-07-31
playhtml: reevaluate cursor render predicates

spencerc99:
## Summary

- Reevaluate every current remote cursor when `shouldRenderCursor` changes.
- Reuse the existing cursor update path so filtered cursors are removed and newly allowed cursors are recreated without waiting for another peer event.
- Add focused regression coverage for both visibility directions and a `playhtml` patch changeset.

## Root cause

`CursorClientAwareness.configure()` updated the stored cursor options, but `shouldRenderCursor` was only read by `updateCursor()`. Existing remote cursor DOM was therefore left in its prior state until that peer published another cursor update.

## Validation

- `bun run test -- src/cursors/__tests__/configure-render-filter.test.ts` (2 tests)
- `bun run test` in `packages/playhtml` (51 files, 501 tests)
- `bunx tsc --noEmit` in `packages/playhtml`
- Real-browser validation with two independent clients: a current remote cursor disappears when the predicate filters it and reappears when the predicate allows it, with no new peer movement.

## Docs and starters

No docs change is needed because the documented `shouldRenderCursor` signature and usage are unchanged. The fix makes `configure()` honor the existing contract immediately. No starter update is needed because imports, options, markup, and CSS are unchanged.

### DIFF
--- a/.changeset/refresh-cursor-render-filters.md
+++ b/.changeset/refresh-cursor-render-filters.md
@@ -1,0 +1,5 @@
+---
+"playhtml": patch
+---
+
+Reevaluate current remote cursors when `shouldRenderCursor` changes so visibility updates do not wait for another peer cursor event.
--- a/packages/playhtml/src/cursors/__tests__/configure-render-filter.test.ts
+++ b/packages/playhtml/src/cursors/__tests__/configure-render-filter.test.ts
@@ -1,0 +1,116 @@
+// ABOUTME: Verifies cursor render filters reevaluate current remote presences.
+// ABOUTME: Predicate changes apply without waiting for another peer update.
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import * as Y from "yjs";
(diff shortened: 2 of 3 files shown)


spencerc99:
## Preview evidence

**Changing shouldRenderCursor immediately hides and restores current remote cursor DOM without another peer update.**

[Open preview](http://127.0.0.1:5173/test/playground.html) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-ms9embkc-9463c0c2) · Commit `fb87f5a`

### Current remote cursor filtered

Current remote cursor filtered

Two clients remain online while the current remote cursor DOM is removed immediately after the predicate changes to false.

[Open captured page](http://127.0.0.1:5173/test/playground.html) · 1280×720

### Current remote cursor restored

Current remote cursor restored

The same current remote cursor reappears immediately after the predicate changes to true, without another peer movement.

[Open captured page](http://127.0.0.1:5173/test/playground.html) · 1280×720

PR #352, opened 2026-08-05
React: keep presence room hooks safe during navigation

### LINKED ISSUE #267: usePresenceRoom can call createPresenceRoom before init during provider remount
## Summary

`usePresenceRoom()` is documented as safe before playhtml initialization, but it can still call `playhtml.createPresenceRoom()` while the core singleton is not ready during a React provider remount or route-transition lifecycle. That throws and can break the whole React island.

Observed stack from `spencers-website` when navigating from `/` to `/creation` through Astro view transitions:

```text
Uncaught Error: playhtml.createPresenceRoom is not available before init()
    at Object.Cg [as createPresenceRoom] (PlayhtmlProvider.CF4Fy-Xg.js:780:4360)
    at PlayhtmlProvider.CF4Fy-Xg.js:780:15189
    at el (client.DWvMSTGb.js:24:24164)
    at $t (client.DWvMSTGb.js:24:42048)
    at ya (client.DWvMSTGb.js:24:34316)
```

When the page is refreshed directly, the route renders correctly. The crash only appeared during client-side navigation.

## Why this looks like a PlayHTML React bug

The hook contract says:

```ts
/**
 * Join a presence room. Safe to call before playhtml has initialized:
 * returns `null` until sync completes. When `name` changes, briefly returns
 * `null` during the transition between rooms.
 */
export function usePresenceRoom(name: string): PresenceRoom | null {
  const { isLoading } = useContext(PlayContext);
  const [room, setRoom] = useState<PresenceRoom | null>(null);

  useEffect(() => {
    if (isLoading) return;
    const r = playhtml.createPresenceRoom(name);
    setRoom(r);
    return () => {
      r.destroy();
      setRoom(null);
    };
  }, [isLoading, name]);

  return room;
}
```

Relevant files in the repo:

- `packages/react/src/hooks.ts:152-171`: `usePresenceRoom` gates only on `PlayContext.isLoading`.
- `packages/react/src/PlayProvider.tsx:166`: `const [hasSynced, setHasSynced] = useState(playhtml.isLoading === false);`
- `packages/playhtml/src/index.ts:1910-1913`: `createPresenceRoom` throws when core `hasSynced` is false.

That creates a mismatch: React context can expose `isLoading: false` from the singleton-derived initial state in a remount scenario, while the core singleton still rejects `createPresenceRoom()` for the current lifecycle. `usePresenceRoom()` then violates its own safe-before-init contract.

## Reproduction shape

The app-side case was:

1. Astro site mounts a top-level `PlayProvider` with `client:only="react"` and `transition:persist`.
2. A child component calls `usePresenceRoom("chat")` during render lifecycle.
3. Navigate from `/` to `/creation` using Astro client navigation.
4. During the route transition/provider remount timing, `usePresenceRoom` effect calls `playhtml.createPresenceRoom("chat")` before core init is available.
5. The exception aborts the island and the creations list does not render.
6. Direct refresh on `/creation` works.

## Expected behavior

`usePresenceRoom(name)` should never throw before init. It should return `null` until the core singleton can safely create the room, then create and return the room after readiness is confirmed.

## Actual behavior

`usePresenceRoom(name)` can throw `playhtml.createPresenceRoom is not available before init()` despite the hook contract saying it is safe before initialization.

## App-side workaround used

In `spencers-website`, I worked around this by delaying components that call PlayHTML APIs until after `hasSynced` has been observed in a committed effect tick. That avoided the navigation crash, but the package should own this safety so app code does not need lifecycle guards around a hook documented as pre-init safe.

Temporary workaround commit there: `b565415 Fix playhtml room timing on navigation`.

## Suggested fix direction

The React package should make `usePresenceRoom` readiness depend on the actual core readiness, not only context state from the provider remount.

Possible approaches:

- Make `PlayProvider` initialize `hasSynced` as `false` on mount and only set it true after the current mount observes `playhtml.ready`/`init()` completion.
- Or have `usePresenceRoom` catch/prevent the core pre-init path by checking actual singleton readiness before calling `createPresenceRoom`.
- Add a regression test where `playhtml.isLoading` starts false for provider initial state but `createPresenceRoom` still throws until the current `ready` promise resolves; `usePresenceRoom` should return `null`, not throw, then create the room after readiness.

The important contract is that consumers should be able to call `usePresenceRoom()` unconditionally in a component rendered under `PlayProvider`, with `null` representing not-ready state.

spencerc99:
## Summary

- Keep `usePresenceRoom()` at `null` when React renders while PlayHTML is temporarily unsynced.
- Retry room creation when `playhtml.ready` resolves or navigation finishes.
- Preserve other `createPresenceRoom()` errors instead of swallowing them.
- Add unit and real-core integration coverage for provider remounts and room-name changes during readiness.

## Root cause

`PlayProvider` can retain a synced context value while core PlayHTML is rebuilding its page room during navigation. If the provider remounts in that window, `usePresenceRoom()` calls `createPresenceRoom()` before core sync completes and throws instead of honoring the hook's documented `null` state.

## Validation

- `bun run test` in `packages/react`: 52 unit tests and 2 real-core integration tests passed.
- `bun run build` in `packages/react` passed.
- Tested Spencer's website against local builds of this branch through `/creation -> /about -> /creation -> / -> /creation`.
- The final and directly loaded `/creation` views each rendered 187 rows with no `createPresenceRoom` error.

The public docs already describe the intended behavior: the hook returns `null` until synced. Starter templates do not use this hook, so neither needs an update.

Fixes #267

### DIFF
--- a/.changeset/polite-presence-rooms.md
+++ b/.changeset/polite-presence-rooms.md
@@ -1,0 +1,5 @@
+---
+"@playhtml/react": patch
+---
+
+Keep `usePresenceRoom()` returning `null` instead of throwing when React renders ahead of PlayHTML readiness during navigation or provider remount timing.
--- a/packages/react/src/__tests__/binding.integration.test.tsx
+++ b/packages/react/src/__tests__/binding.integration.test.tsx
@@ -1,10 +1,10 @@
-// ABOUTME: Tests React element binding lifecycle against the real playhtml core.
-// ABOUTME: Verifies data-source changes remove handlers created after asynchronous ID assignment.
+// ABOUTME: Tests React hooks and element bindings against the real playhtml core.
+// ABOUTME: Verifies binding cleanup and presence-room readiness across navigation.
 import React from "react";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
-import { cleanup, render, waitFor } from "@testing-library/react";
+import { act, cleanup, render, waitFor } from "@testing-library/react";
 import { elementHandlers, playhtml, resetPlayHTML, TagType } from "playhtml";
-import { withSharedState } from "../index";
+import { PlayProvider, usePresenceRoom, withSharedState } from "../index";
 
 describe("CanPlayElement binding lifecycle", () => {
   beforeEach(async () => {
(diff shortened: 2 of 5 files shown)


spencerc99:
## Code review

PR #358, opened 2026-08-07
Fix Firefox presence message handling

spencerc99:
## Summary

- Receive PartySocket messages and lifecycle events through direct handler properties when that interface is available.
- Keep the event-listener path for injected sockets that do not expose the complete handler interface.
- Add regression coverage and a Firefox extension release note.

## Root cause

Firefox WebExtension content scripts can silently skip listeners on custom `EventTarget` subclasses. The native WebSocket received presence events and PartySocket handled them, but PartySocket's redispatched events never reached PlayHTML's listeners. Remote peers therefore never entered the shared presence store, leaving the presence control hidden. The skipped `open` event also left the connection grace timer active, which could mark a working connection unreachable.

PartySocket calls its `onmessage`, `onopen`, `onclose`, and `onerror` properties directly before calling `dispatchEvent()`. Using those handlers bypasses Firefox's broken redispatch path.

## User impact

Firefox users can see remote cursors and the bottom-right presence and chat control when other people are on the page. Working connections no longer emit a false presence-degraded error after the grace period.

## Validation

- Verified the production-style Firefox build on Wikipedia after the 15-second connection grace period; the presence control remained visible.
- Confirmed the Firefox page console contained no `presence transport unreachable` or `realtime presence degraded` errors.
- `packages/playhtml`: 503 tests passed.
- PlayHTML package build passed.
- Firefox extension build passed.
- Focused transport suite: 15 tests passed, including direct `message`, `open`, `close`, and `error` handlers and guarded cleanup.

## Documentation

No public documentation or starter-template changes are needed. This fixes internal transport compatibility without changing the public API or usage.

### DIFF
--- a/.changeset/firefox-presence-messages.md
+++ b/.changeset/firefox-presence-messages.md
@@ -1,0 +1,5 @@
+---
+"playhtml": patch
+---
+
+Receive realtime presence messages reliably in Firefox so remote cursors and presence controls appear.
--- a/extension/PENDING.md
+++ b/extension/PENDING.md
@@ -2,6 +2,7 @@
 
 - Open History from every new tab or the popup to see time by site, active explorations, smaller places you returned to, daily cursor portraits, and a moving landscape of real browsing traces.
 - Browse a full year of History without repeated page metadata overwhelming the extension.
+- Presence controls and remote cursors now appear reliably in Firefox.
 
 <!--
 Add a bullet here in any PR that touches extension/**. The release-prep workflow
--- a/packages/playhtml/src/__tests__/presence-transport.test.ts
+++ b/packages/playhtml/src/__tests__/presence-transport.test.ts
@@ -61,11 +61,28 @@ class FakeSocket {
   }
 }
 
-class PropertyMessageSocket extends FakeSocket {
+class PropertyHandlerSocket extends FakeSocket {
   onmessage: ((event: MessageEvent) => void) | null = null;
+  onopen: ((event: Event) => void) | null = null;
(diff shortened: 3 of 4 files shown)


spencerc99:
## Preview evidence

**Firefox receives presence and lifecycle events without falsely marking the live connection unreachable.**

[Open preview](https://en.wikipedia.org/wiki/Main_Page) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-msjcst92-0de9557b) · Commit `5026487`

### Firefox presence control after lifecycle fix

Firefox presence control after lifecycle fix

After the connection grace period, Firefox still renders the live presence control and the console contains no false presence-unreachable error.

[Open captured page](https://en.wikipedia.org/wiki/Main_Page) · 1280×887

spencerc99 (commented):
## Code review

[P2] The Firefox workaround only covers `message`; `open`, `close`, and `error` still use the same broken custom-`EventTarget` redispatch path in `packages/playhtml/src/presence-transport.ts:126`.

PartySocket calls each `on*` property directly and then calls `dispatchEvent()`. Firefox WebExtension content scripts skip listeners on this custom `EventTarget`, so `onOpen` never clears the 15-second unreachable timer and a working Firefox connection can still be marked degraded. Close/error reconnect accounting also remains inactive.

Use the direct `onopen`, `onclose`, and `onerror` properties alongside `onmessage`, with the same guarded cleanup, and extend the regression test to cover those lifecycle handlers.

PR #362, opened 2026-08-11
extension: reduce background performance overhead

spencerc99:
## Summary

- maintain bounded daily activity and recent-domain aggregates as events are stored
- read milestone data from those aggregates instead of repeatedly scanning raw cursor and navigation history
- batch routine extension storage writes once per second while keeping clicks immediate
- defer cursor selectors and computed styles until they are needed

## Why

Chrome reported that WWO was using substantial resources. Controlled traces identified cursor collection as the largest active-page JavaScript cost. The milestone alarm also replayed raw history every five minutes, making an auxiliary feature progressively more expensive as local history grew.

The aggregate fields add bounded storage: one daily global activity record, up to five visit timestamps per domain, and one favicon URL capped at 2,048 characters. Existing installations begin daily counters after updating and do not scan old raw events to backfill them.

## Performance

In the same three-run Chrome interaction trace, mean extension ScriptDuration changed from 51.6 ms to 40.7 ms, a 21% reduction. TaskDuration remained noisy, so this does not claim an equivalent end-to-end improvement.

## Validation

- 104 extension test files, 618 tests
- `bun run build-packages`
- production Chrome MV3 extension build
- unpacked build synchronized and checksum-verified
- `git diff --check`

The repository-wide type-check still reports unrelated errors in untouched extension UI, shared visualization, and Worker files. None of the reported errors are in this PR's files.

### DIFF
--- a/extension/PENDING.md
+++ b/extension/PENDING.md
@@ -6,6 +6,7 @@
 - Use the extension in Safari 18 or later on Mac.
 - Turn on Safari website access once during setup, then close the setup tab when you’re done.
 - Get clear recovery steps if Safari cannot save your setup choices.
+- Reduced WWO's impact on browser performance during long browsing sessions.
 
 <!--
 Add a bullet here in any PR that touches extension/**. The release-prep workflow
--- a/extension/PENDING.md
+++ b/extension/PENDING.md
@@ -7,6 +7,7 @@
 - Turn on Safari website access once during setup, then close the setup tab when you’re done.
 - Get clear recovery steps if Safari cannot save your setup choices.
 - Reduced WWO's impact on browser performance during long browsing sessions.
+- Shortened browsing portrait captions and kept long site names inside their cards.
 
 <!--
 Add a bullet here in any PR that touches extension/**. The release-prep workflow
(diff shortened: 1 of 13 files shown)


spencerc99:
## Code review

- **P2 [`extension/src/collectors/CursorCollector.ts:92`](https://github.com/spencerc99/playhtml/blob/34c3efbe6a5d362f8a441d1b94317d4ad92f37a6/extension/src/collectors/CursorCollector.ts#L92) leaves cursor style undefined until the pointer crosses an element boundary.** The new code assigns `currentCursorStyle` only from `mouseover`, but the collector is normally enabled while the pointer is already over the page. Moving within that same element runs only the `mousemove` path, so every real-time and archived move is emitted with `cursor: undefined` until the pointer enters another element; on a full-page canvas that can last the whole session. Preserve the reduced style reads by initializing the style on the first `mousemove` when `lastCursorStyleTarget` is still null, then relying on `mouseover` for later target changes.

- **P2 [`extension/src/history/walkingRecord.ts:998`](https://github.com/spencerc99/playhtml/blob/34c3efbe6a5d362f8a441d1b94317d4ad92f37a6/extension/src/history/walkingRecord.ts#L998) adds a user-facing caption change without the required finished-surface screenshots.** This head changes the visible walking-record caption and clamps it in the card, but the current PR description and comments contain no screenshot evidence. The repository instructions require screenshots of every user-facing extension change. Capture the actual walking-record card with the shortened/clamped caption and add it to the PR before merge.

PR #367, opened 2026-08-13
Improve Wikipedia live presence

spencerc99:
## Summary

- keep one Wikipedia article-name stable across every open extension tab
- show other readers' active text selections in real time using their cursor colors
- count each connected identity once in the "elsewhere" total, including readers without a current page payload
- preserve the full query string for Wikipedia `/w/index.php` editing rooms while keeping normal article tracking parameters normalized

## Why

Wikipedia chat names could race during simultaneous tab startup, live selections were not visible to peers, and the lobby count omitted connected people when their page metadata was missing or stale. Editing URLs also collapsed distinct query-based sessions into the same room.

## Implementation

The background worker now owns article-name generation and persistence, which serializes simultaneous initialization and broadcasts storage changes to open tabs. Wikipedia selections use presence data containing DOM paths and offsets only. Remote selections render through the CSS Custom Highlight API and disappear when the selection or connection ends.

The presence pill counts lobby identities independently from jump destinations. A fresh, visible page payload is required only to show the jump button.

Wikipedia routes with `action=edit` or `veaction` preserve the full query string in the room key. This covers both `/w/index.php` and `/wiki/Title` editor URLs without fragmenting ordinary reading rooms. Selection updates publish at most every 250ms and unchanged ranges are not sent again.

## Validation

- `bun run build-packages`
- full extension suite: 105 files, 630 tests
- focused review suite: 4 files, 43 tests
- `bunx tsc --noEmit`
- Chrome MV3 production build
- Firefox MV2 production build
- generated manifests still declare `chrome_url_overrides.newtab = newtab.html` for both browsers
- `git diff --check`

## Screenshots

Pending live installed-extension verification. Chrome is not connected to Codex, so the required multi-reader selection and presence screenshots could not be captured yet.

### DIFF
--- a/extension/PENDING.md
+++ b/extension/PENDING.md
@@ -14,3 +14,5 @@ extension/website/public/changelog/media/ and reference them here:
 ![Screenshot title](/changelog/media/file.png)
 ![video: Demo title](/changelog/media/file.mp4)
 -->
+
+- Wikipedia now keeps your article-name consistent across tabs, shows other readers' live text selections in their cursor colors, keeps editing sessions separate by URL, and counts each connected reader once across pages.
--- a/extension/PENDING.md
+++ b/extension/PENDING.md
@@ -15,4 +15,6 @@ extension/website/public/changelog/media/ and reference them here:
 ![video: Demo title](/changelog/media/file.mp4)
 -->
 
-- Wikipedia now keeps your article-name consistent across tabs, shows other readers' live text selections in their cursor colors, keeps editing sessions separate by URL, and counts each connected reader once across pages.
+- Wikipedia now keeps your article-name consistent across tabs and counts each connected reader once across pages.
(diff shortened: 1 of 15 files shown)


spencerc99:
## Code review

- **P2 [`extension/src/custom-sites/wikipedia.ts:381`](https://github.com/spencerc99/playhtml/blob/5e31f4e38bd3816c15eeb5bf7786cd7cd92cbf50/extension/src/custom-sites/wikipedia.ts#L381) enables user-facing live selections without the required finished-surface screenshots.** The PR body explicitly says installed-extension verification is still pending, and its comments contain no screenshots. The repository instructions require every user-facing extension change to be verified on the real affected surface and accompanied by screenshots of the key states. Capture the installed multi-reader selection and presence-pill states and add them to the PR before merge.

PR #372, opened 2026-08-13
partykit: prevent writes before room hydration

spencerc99:
## Summary

- Retry Supabase document hydration up to three times before entering transient mode.
- Abort timed-out reads so stale requests cannot overlap later attempts.
- Derive every shared-data write decision from one room state: `quarantined`, `loading`, `transient`, `save-paused`, or `ready`.
- Route every document database write through `saveDocumentBase64`, including autosave, compaction, snapshot restore, hard reset, and `admin/force-save-live`.
- Release reset save locks with `try/finally` instead of a one-second timer.
- Keep awareness updates available while Yjs document writes are read-only.
- Keep restored rooms writable after an authoritative snapshot replaces the live document.

## Why

The `playhtml.fun-fridge` incident showed a persisted document shrinking from a normal snapshot to a nearly empty snapshot after a Supabase load timeout. The room entered transient mode correctly, but a later lifecycle callback marked persistence available without proving that hydration completed. A delayed autosave then persisted the empty or partial in-memory document.

This change makes hydration completion an independent, required condition for writes. Persistence recovery alone cannot reopen autosave, Yjs updates, bridge writes, compaction, or admin mutations.

The write guard now also lives at the database boundary. Callers cannot bypass quarantine or hydration checks by issuing a direct document upsert. The only exception is an explicit quarantine-repair restore, which must be able to replace the document that caused the quarantine.

## Failure coverage

| Failure case | Expected behavior |
| --- | --- |
| First or second Supabase read fails | Retry with bounded exponential delay. |
| Final configured read succeeds | Hydrate normally and leave transient mode disabled. |
| All reads fail | Enter transient mode after the final attempt. |
| A read exceeds its timeout | Abort that attempt before retrying. |
| Persistence later reports available without hydration | Keep every shared-data write path closed. |
| A delayed autosave runs after failed hydration | Perform zero Supabase upserts and preserve the stored row. |
| A client sends a Yjs update before hydration | Reject the document update. |
| A client sends awareness before hydration | Apply and broadcast awareness normally. |
| Bridge subscribe, apply, flush, registration, or URL metadata runs before hydration | Reject or skip the write without changing room state. |
| Automatic compaction runs before hydration | Skip compaction. |
| An admin mutation runs before hydration | Return 503 without persisting the live document. |
| `admin/force-save-live` runs while quarantined | Return 503 and preserve the protected document. |
| A caller invokes the document save function before hydration | Reject the write before Supabase is called. |
| Autosave or an admin write runs during a reset | Treat the room as `save-paused`, perform zero upserts, and preserve the stored row. |
| Snapshot restore succeeds | Release the save lock immediately and reopen writes. |
| Snapshot restore database write fails | Release the save lock in `finally` and propagate the failure. |
| Hard reset succeeds | Release the save lock immediately after installing the authoritative document. |
| Hard reset database write fails | Release the save lock in `finally` and preserve the previous document. |
| A quarantined snapshot restore succeeds | Use the explicit repair operation, install the authoritative document, clear recovery state, and reopen writes. |
| Snapshot restore succeeds but quarantine cleanup fails | Report the durable restore while leaving the room quarantined and read-only. |

## Database write boundary

`saveDocumentBase64` is the only production `documents.upsert` call. It enforces the room state before writing and records the persisted snapshot only after Supabase succeeds.

| Operation | Allowed room state |
| --- | --- |
| Autosave and `admin/force-save-live` | `ready` |
| Compaction, hard reset, and normal snapshot restore | `ready` or `save-paused` |
| Explicit quarantine-repair restore | Any state, because it replaces the document that caused the write park |

## Validation

- `bun test partykit/__tests__` (234 passed, 1,220 assertions)
- `bunx tsc -p partykit --noEmit`
- `bunx wrangler deploy --config partykit/wrangler.jsonc --env="" --dry-run --outdir <temp-dir>`
- `git diff --check origin/wip/transient-mode-retries...HEAD`
- Source audit: one production `documents.upsert`, inside `saveDocumentBase64`

No public API, package, docs, or visual UI changed. No deployment was performed.

Browser-side offline persistence is not included. Adding IndexedDB requires a separate design for reset epochs and stale-client reconciliation.

### DIFF
--- a/partykit/__tests__/persistenceMode.test.ts
+++ b/partykit/__tests__/persistenceMode.test.ts
@@ -128,6 +128,8 @@ describe("formatPersistenceFailureLog", () => {
     expect(message).toContain("attempts=3");
     expect(message).toContain("connection timeout");
     expect(message).toContain("TRANSIENT MODE");
+    expect(message).toContain("awareness may continue");
+    expect(message).toContain("shared-data writes disabled");
     expect(message).toContain("autosave disabled");
   });
 });
@@ -147,7 +149,7 @@ describe("createPersistenceUnavailableResponse", () => {
     expect(body).toEqual({
       error: "persistence_unavailable",
       message:
-        "Supabase persistence is unavailable for this room; admin writes are disabled while realtime runs in transient mode.",
+        "Supabase persistence is unavailable for this room; shared-data and admin writes are disabled while awareness runs in transient mode.",
       roomId: "example-room",
       failedAt: "2026-05-26T21:05:45.000Z",
       reason: "connection timeout",
--- a/partykit/__tests__/persistenceMode.test.ts
+++ b/partykit/__tests__/persistenceMode.test.ts
@@ -80,6 +80,27 @@ describe("retryWithTimeout", () => {
     expect(failures).toEqual([1]);
   });
 
+  test("can recover on the final configured attempt", async () => {
+    const attempts: number[] = [];
+
+    const result = await retryWithTimeout(
+      (_signal, attempt) => {
+        attempts.push(attempt);
+        if (attempt < 3) throw new Error(`failure ${attempt}`);
+        return Promise.resolve("loaded");
+      },
+      {
+        attempts: 3,
+        timeoutMs: 100,
+        retryDelayMs: 1,
+        errorMessage: "timed out",
+      }
+    );
+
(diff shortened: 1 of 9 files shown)


spencerc99:
## Code review

PR #375, opened 2026-08-15
Stabilize movement audio playback

spencerc99:
## Summary

- reuse faded trail voices instead of replacing them after short inactive gaps
- normalize movement gain against audio time so delayed frames do not create full-volume spikes
- release archive voices on silent frames and disconnect playback graphs after resets
- add regression coverage for repeated gap/resume cycles, delayed frames, silent archive playback, and reset cleanup
- add the user-facing extension release note

## Root cause

Short inactive gaps stopped and forgot a trail's oscillators. If the trail resumed before the fade completed, the sound engine created replacement oscillators while the previous graph remained connected. Repeated gaps accumulated overlapping audio graphs over time.

Archive playback also skipped sound-engine ticks when no trail was active and cleared stopped voices without disconnecting their graph. Main-thread delays amplified the problem because cursor distance was treated as per-frame velocity regardless of how long the frame took.

## Impact

Live portrait and archive sound remains stable through playback slowdowns, inactive gaps, and longer sessions. The fix applies to the common audio path, independent of the optional chord, cursor-instrument, and crossing settings.

## Validation

- `bun run build-packages`
- `bun run --cwd extension test --run` (642 tests)
- focused audio and trail lifecycle tests (11 tests)
- `bun run lint`
- `bun run --cwd extension/website build`
- prolonged local browser runs on `/portrait/` and `/archive/` with sound enabled and no console warnings or errors

This is an audio-only behavior change. Screenshots do not demonstrate the result, and the affected portrait contains live browsing-activity traces, so no public media was uploaded.

### DIFF
--- a/extension/PENDING.md
+++ b/extension/PENDING.md
@@ -18,4 +18,3 @@ extension/website/public/changelog/media/ and reference them here:
 - Wikipedia now keeps your article-name consistent across tabs and counts each connected reader once across pages.
 - Wikipedia shows other readers' live text selections in their cursor colors.
 - Wikipedia editing sessions now stay separate from article-reading rooms.
-- Sound on live and archive portraits now stays smooth during playback slowdowns and longer sessions.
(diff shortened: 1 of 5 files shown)


spencerc99:
## Code review

PR #379, opened 2026-08-15
Remove test-only changeset

spencerc99:
## Summary

Remove the changeset added by PR #377.

## Why

PR #377 only added regression-test coverage and did not change package behavior. Publishing a `playhtml` patch for test-only changes would create an unnecessary release.

## Validation

- Confirmed the branch diff contains only the changeset deletion
- `git diff --check`


spencerc99:
Closing because the merged regression-coverage work does not need a follow-up release change.

PR #381, opened 2026-08-16
Fix live portrait trail lifecycle

spencerc99:
## Summary

- Wait eight seconds after the draw head reaches the tip before a trail begins settling.
- Keep settled history dim while a resumed suffix draws at the live opacity.
- Keep settled trails visible for 60 seconds, then fade them out over 8 seconds.
- Keep click and hold ripples visible with their owning trail and apply the trail's departure fade to them.
- Give each replacement cursor segment its own render identity.
- Preserve previously drawn geometry when a live batch adds points.
- Subdivide sparse live geometry into segments no longer than 20 px and allocate at least 32 ms per rendered segment.
- Cap hold ripple size and expansion duration at 3×.
- Prevent removed trails from being rebuilt from events that remain in the live buffer.

## Root cause

The settle timer started when the last point arrived. If a long trail needed more than eight seconds to finish drawing, it settled as soon as the draw head reached the tip. Later points reused the same dimmed SVG path, so the new portion also appeared faded.

Growing live paths were rounded and resampled from the complete point list after every batch. Appending points could therefore reshape geometry that was already visible. Draw progress also advanced by rendered point index. Sparse input could leave hundreds of pixels between two rendered points, and the 30-second duration cap could override point-level pacing on very long trails. Those cases made large portions appear in one step even when the total duration looked reasonable.

Live paths now keep a stable geometric prefix. Sparse segments are subdivided independently, so appending a point does not alter existing geometry. Geometry-driven duration is not capped, which guarantees the spatial and per-segment pacing floors for long trails. Only long idle time in the captured timestamps remains capped.

The path renderer now uses a dim base path and a separate live suffix. The suffix fades into the base when it settles. Click and hold ripples remain mounted with their owning trail and inherit that trail's visibility transition. Hold expansion is bounded so malformed durations cannot produce an unbounded ripple.

## User impact

Live movement draws progressively instead of appearing as a large completed section. The newest part remains at full opacity until the trail settles. Resumed movement adds a bright suffix without relighting settled history. Finished trails and their ripples fade together instead of disappearing independently.

## Validation

- `vitest run` in `extension/website` (25 files, 115 tests)
- `tsc --noEmit` in `extension/website`
- `vite build` in `extension/website`
- Regression coverage for prefix-stable subdivision, spatial pacing, per-segment pacing, and trails longer than the 30-second timestamp cap
- Deterministic browser fixture: a 4-second trail remained fully opaque until 12 seconds, then settled over 1.2 seconds
- Deterministic browser fixture: resumed history remained dim while the live suffix rendered at full opacity
- Deterministic browser fixture: a 20-minute hold was capped at 3×
- Final Cloudflare preview observed for 128 seconds with stable trail identity matching: no multi-hundred-pixel active-head arrivals; largest sample was 95 px over 134 ms
- Final Cloudflare preview: 956 samples captured while ripple wrappers were fading with trails, with zero ripple-only fade samples while the owning trail remained fully visible

### DIFF
--- a/extension/website/shared/components/ClickRipple.tsx
+++ b/extension/website/shared/components/ClickRipple.tsx
@@ -1,6 +1,6 @@
 // ABOUTME: Shared ripple effect for click/hold visualization
 // ABOUTME: Used by AnimatedTrails and AnimatedClicks so ripple logic stays DRY
-import { useState, useEffect, useMemo, useRef, memo } from "react";
+import { useState, useEffect, useRef, memo } from "react";
 import { ClickEffect } from "../types";
 
 export interface RippleSettings {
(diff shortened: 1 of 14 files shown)


spencerc99:
## Preview evidence

**Live trail extensions stay bright over dim history, and click/hold ripples fade independently through trail departure.**

[Open preview](https://cx-fix-website-portrait-trai.we-were-online-website.pages.dev/portrait/) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-mswgrmm2-fee3a848) · Commit `ce38d89`

### Bright live extension over settled trail

Bright live extension over settled trail

The settled portion remains dim while only the resumed suffix draws at full color.

[Open captured page](http://127.0.0.1:5173/portrait-lifecycle-debug/) · 1280×720

### Ripple completes after its trail departs

[Watch the demo video](https://pr-evidence.spencerc99.workers.dev/media/ev-mswgrmm2-fee3a848/ripple-completes-after-its-trail-departs/ripple-completes-after-its-trail-departs.mp4)

A capped long-hold ripple continues fading to zero after its parent trail has left the canvas.

[Open captured page](http://127.0.0.1:5173/portrait-lifecycle-debug/) · 1280×720

spencerc99:
## Code review

- **P2 [Add the required finished-state screenshots](https://github.com/spencerc99/playhtml/blob/2e5fe811cbc5e48e8b03252b7d5c8c3d8b7b2fed/AGENTS.md#L121).** This changes user-visible portrait trail behavior, but the PR body and evidence comment contain only an MP4. The applicable instruction requires screenshots for every user-facing change and the key states of an interactive flow. Attach still screenshots showing the settled trail, the departure fade, and the resumed trail state before merge.

PR #382, opened 2026-08-17
worker: make event exports stable during ingestion

spencerc99:
## Summary

- Replace offset pagination with ascending `(ts, id)` keyset pagination.
- Report `eventCount` from the rows included in the export.
- Cover delayed inserts and equal-timestamp ordering with behavior-level tests.

## Root cause

The export used positional offsets while event ingestion remained active. Inserting a row before a later page shifted an event from the first page onto the second page. A reproduction exported 1,002 rows with only 1,001 unique IDs.

## Impact

Admin exports no longer duplicate or skip rows from the original result set when delayed events are inserted before the current cursor. The response format is unchanged.

## Testing

- `bun run test:extension-worker` (80 tests)
- `bun run check:extension-worker`
- `bun run doctor`
- `bun run smoke:extension-worker`
- `git diff --check`


spencerc99:
## Code review

PR #383, opened 2026-08-17
extension: preserve current-day milestone activity

spencerc99:
## Summary

- Preserve the newest local day's compact milestone activity when older events finish saving later.
- Add a two-batch regression test and an extension release note.

## Root cause

The compact milestone aggregate reset whenever an incoming event had a different local day. Event batches from separate tabs can arrive out of timestamp order, so a late previous-day batch replaced the current day's accumulated distance and screen time.

## Impact

Daily distance and screen-time milestone progress no longer disappears when older browsing activity arrives after newer activity.

## Testing

- `bun build-packages`
- `cd extension && bun run test -- src/__tests__/LocalEventStore.test.ts` (40 passed)
- `cd extension && bun run build`
- `git diff --check`
- `bun run check:extension` (blocked by the existing ref type error at `extension/src/entrypoints/commute/commute.tsx:1048`, outside this diff)

Screenshots are not applicable because this change does not alter the UI.


spencerc99:
## Code review

PR #386, opened 2026-08-17
Detect JavaScript files that shadow TypeScript sources

spencerc99:
## Summary

- Ignore TypeScript-emitted `.js` and `.js.map` siblings under `packages/*/src` and `extension/src`.
- Make `bun run doctor` detect untracked `.js` files that shadow tracked `.ts` or `.tsx` sources, including files hidden by `.gitignore`.
- Cover ignored shadow files and non-shadowing JavaScript with a real temporary Git repository fixture.

## Root cause

A stray TypeScript emit can place JavaScript beside source files. Module resolution may select that JavaScript instead of the TypeScript source, which caused the package build to read `packages/playhtml/src/index.js` and fail with `"playhtml" is not exported by "src/index.js"`.

The existing doctor checked generated dependencies and build artifacts only. It did not inspect source trees, and ordinary untracked-file checks would stop seeing the emitted files after adding ignore rules.

## Behavior

The source-tree ignore rules are limited to `packages/*/src` and `extension/src`. The doctor queries both ordinary and ignored untracked JavaScript, then reports a file only when Git tracks the matching TypeScript source. Each failure names the emitted file and tells the developer to remove it.

No tracked JavaScript currently exists in these source trees. This is a development guardrail only, so no changeset, public documentation update, starter update, or screenshot is needed.

## Verification

- Reproduced the missing guardrail with an untracked `packages/playhtml/src/index.js`; the previous doctor output did not report it.
- Verified the updated doctor reports the same file after `.gitignore` hides it.
- `bun run setup`
- `bun run test:doctor`
- `bun run doctor`
- `bun run lint`
- `git diff --check`
- Conservative review of `origin/main...HEAD`: no findings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,12 @@ site-dist
 dist-ssr
 *.local
 
+# TypeScript emit beside source files
+packages/*/src/**/*.js
+packages/*/src/**/*.js.map
+extension/src/**/*.js
+extension/src/**/*.js.map
+
 # Editor directories and files
 .vscode/*
 !.vscode/extensions.json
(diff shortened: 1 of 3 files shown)


spencerc99:
## Code review

PR #387, opened 2026-08-17
Speed up extension browser history loading

spencerc99:
## Summary

- Reopen browser-history records viewed in the last 15 minutes without recomputing them.
- Keep stale cached records visible while refreshing them in the background.
- Retry a stale cached record when navigation interrupts its background refresh.
- Render the base record before cursor paths finish, with skeletons confined to browsing portraits and the movement landscape.
- Restore favicons for older records while scanning only the selected period.
- Keep an explicit movement-loading preview at `walking-record.html?previewLoading=movement`.
- Tighten hour-and-minute totals in “how you browsed,” such as `1h20m`.

## Root cause

The page discarded completed records when a new tab closed. Every new tab repeated the full history scan and cursor-path reconstruction, even when the same period had been shown moments earlier.

The base history scan feeds all record sections. Cursor-path reconstruction only enriches browsing portraits and the movement landscape, but the page previously treated that work as one top-level loading phase.

Domain aggregates now store each domain’s latest favicon. Records created before that field was added can still lack icons, so the bounded selected-period scan also returns the latest valid favicon it encounters for each visited domain.

In a synthetic 24,000-event benchmark, the isolated favicon phase dropped from 10.05 seconds to 1.2 milliseconds. This is not an end-to-end browser measurement. A fresh cache hit skips record recomputation entirely for 15 minutes.

## Validation

- `bun run build-packages`
- `bun run check:extension`
- `bun run test:extension` (116 files, 701 tests)
- `bun run build-extension`
- `git diff --check`
- Verified the built Chrome MV3 artifact matches the copy in the main extension build directory.
- Verified the loading preview on the installed production extension. Portrait and movement skeletons remain stable, domain favicons render, and the rest of the record stays available.

## Screenshots

A finished-state screenshot was captured from the installed extension but not uploaded because it contains local browsing-domain data. The Chrome browser bridge does not capture `chrome-extension://` pages, so the local Mac capture includes the browser chrome and visible domains.


spencerc99:
## Code review

- **P2 [`extension/src/entrypoints/newtab/newtab.tsx:104`](https://github.com/spencerc99/playhtml/blob/53f538eeb7be237056a494c1a787108bb8589b2f/extension/src/entrypoints/newtab/newtab.tsx#L104) can strand a stale cached record after an interrupted refresh.** A stale cache hit is inserted into `records` at lines 128–138 before its background refresh finishes. If the user changes periods, cleanup cancels that refresh at lines 216–218; returning to the period then finds `records[recordKey]` and exits at lines 104–108, so the stale record is never refreshed again during this page lifetime. Keep stale cache state distinguishable from a completed in-memory record (or resume refresh on revisit), and add navigation/cancellation coverage.

- **P2 [`extension/src/components/WalkingRecord.tsx:551`](https://github.com/spencerc99/playhtml/blob/53f538eeb7be237056a494c1a787108bb8589b2f/extension/src/components/WalkingRecord.tsx#L551) changes the shipped new-tab UI without the required screenshots.** The new inline loading states (also line 603) and compact duration presentation are user-facing, while the PR body explicitly says screenshots are unavailable. Repository instructions require screenshots of the finished affected surface and the key interactive states, and require stopping when capture is blocked. Attach evidence for the base record, movement-loading states, and final populated state before merge.

PR #388, opened 2026-08-17
extension: stop overriding the new tab page, add top nav to popup

spencerc99:
## Summary

Three related popup/navigation changes prompted by user churn after the new-tab takeover (~100 uninstalls):

1. **Revert the new tab override.** The walking-record history page no longer replaces the browser's new tab page. Removing `chrome_url_overrides` from `wxt.config.ts` was not enough — WXT auto-registers any entrypoint directory named `newtab`, so the entrypoint is renamed to `walking-record/` (served as `walking-record.html`, all references updated). A regression test asserts the built manifest never grows the override back.

2. **Hide the time page.** The `stats.html` per-domain time view is unlinked from the standalone page nav and the popup. The page itself stays live at its URL. Its "where did my time go" ranking overlaps enough with the new history page that it doesn't earn top-level placement.

3. **New popup header with top nav.** The popup header is now two tight rows: wordmark + cursor identity card, then mono nav links (`portrait ↗ history ↗`, scraps appears when its flag ships) directly below. The subtitle is removed; header height drops 115px → 56px. The presence count moved to the right end of the nav row. The old buried link row under the preview card is gone; changelog moved to the footer. Main sections reordered so the portrait preview card leads and collection status follows. Back links on Profile / Collections / Inventory now stack above the title consistently.

Design was explored through ~10 measured variants in `components-preview` (one-row integrated layouts genuinely cannot fit 3 items at 350px with the full wordmark — the chosen two-row layout fits 3+ items with 108px slack). The preview keeps a single living frame of the final header at 2 and 3 items.

## Testing

- `bun run test:extension`: 669 tests passing (2 new: header/nav structure, section order; 1 new: manifest has no `chrome_url_overrides`)
- `bun run check:extension` / `check:extension-website`: clean
- Rebuilt from wiped `dist/` and verified the manifest contains no new-tab override and emits `walking-record.html`

Screenshots incoming as a follow-up comment (real rendered popup).

## Release notes

Two bullets added to `extension/PENDING.md` (new tab restored; popup nav + reorder).

## Non-applicable checklist items

- No changeset: nothing under `packages/` changed (extension-only).
- No `apps/docs/` audit needed: extension surfaces are not covered by the library docs.
- Starter templates untouched: no core API change.

## Note for review

An announcement postcard CTA changed from `newtab.html` to `walking-record.html` — if any shipped announcement state is keyed on that URL, flag it.

PR #391, opened 2026-08-17
extension: opt-in new tab takeover with an onboarding step

spencerc99:
## Summary

Stacked on #388 (merge that first). Makes the history-page new-tab takeover **opt-in**, replacing the forced manifest override that drove ~100 uninstalls.

**Mechanism:** no `chrome_url_overrides` — a background `tabs.onCreated` listener redirects `chrome://newtab` to `walking-record.html` only when `newtab_takeover_enabled` (browser.storage.local) is true. The pref is cached in the worker (refreshed via `storage.onChanged`), never awaited on the hot path. Redirect replaces history, so Back never bounces through the empty new tab. Measured flash before the page lands: 124–149ms (~130ms avg).

**Two controls, one storage key:**
- A dedicated onboarding step ("See your browsing evolve") with a framed, populated preview of the history page, a default-checked "make this my new tab" box, and one Continue. The key is written explicitly on continue either way — checked → true, unchecked → false; no implicit default.
- A "make this my new tab" checkbox on the walking-record page itself for post-setup control. Verified end-to-end that the setup choice reads back correctly there.

The preview image is a real screenshot of the walking-record page fed with a seeded week of mock events through the extension's own event store (175KB PNG bundled at `extension/public/setup/walking-record-preview.png`). Note: it's static — regenerate via the capture script if the history page is redesigned.

**Bugfix that fell out:** the setup and settings collector lists showed a bare `element` row (no icon, no description) for the flag-gated scraps collector — visible to production users today. The lists now filter by the same released-or-dev-toggle rule the content script uses, and setup only writes collection-mode keys for visible collectors.

## Testing

- `bun run test:extension`: 674 passing (new: setup-step accept/decline storage writes, collector-type gating; updated: step-sequence tests)
- `bun run check:extension` clean
- Both toggle paths verified in a real Chromium against the built extension (storage value + history-page checkbox sync)
- Omnibox focus after redirect not yet verified by automation — checked manually

## Release notes

Three PENDING.md bullets: the setup choice, the merged new-tab wording from #388, and the element-row fix.

## Non-applicable checklist items

- No changeset: extension-only, nothing under `packages/`.
- No docs audit / starter changes: no core API touched.

Screenshots in a follow-up evidence comment.


spencerc99:
## Preview evidence

**Opt-in new tab takeover: onboarding step with live-data preview, grandfathering for existing users, collector-row fix. Shot from the built extension.**

[Open preview](https://github.com/spencerc99/playhtml/pull/391) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-msxspowj-67d75f96) · Commit `b340c2c`

### New setup step (default on)

New setup step (default on)

Framed preview is a real screenshot of the history page seeded with a week of mock data; checkbox arrives checked, Continue writes the choice explicitly either way.

### Setup step, opted out

Setup step, opted out

Unchecking declines; the flow still advances and writes false.

### Bug: leaked element row (before)

Bug: leaked element row (before)

Production setup showed a bare fifth collector row for the flag-gated scraps collector — no icon, no description.

### Configure step (fixed)

Configure step (fixed)

Collector list now filters by the same released-or-dev-toggle rule the content script uses; four rows.

### History page in sync

History page in sync

The walking-record checkbox reads back checked after accepting during setup — both controls share one storage key.

spencerc99:
## Code review

- [P2] Do not drop the wake-up tab while the opt-in preference loads — `extension/src/features/newtab/takeover.ts:36-53`. In Manifest V3, a suspended service worker can be started by the same `tabs.onCreated` event handled here. `enabled` begins as `false`, storage hydration is asynchronous, and the listener returns before that promise resolves, so the first new tab after each worker suspension stays on the browser default even for opted-in users. Make the handler wait for a shared preference-readiness promise (then keep using the cache for later events), or otherwise ensure the event that wakes the worker is evaluated after hydration.

spencerc99:
Addressed the P2: the onCreated handler now awaits the initial storage hydration when it fires before the cache is ready (the service worker wake-up case), then keeps using the in-memory cache for subsequent tabs. Covered by new tests in `newtab-takeover.test.ts`, including the wake-race with a deferred storage read.

PR #392, opened 2026-08-17
Eliminate presence-tick render storm in @playhtml/react

spencerc99:
## Summary

Every cursor-presence tick re-rendered **every playhtml component on the page**. On a room with ~3,000 elements (the production fridge), a one-second drag caused ~340,000 React renders; after this fix the same drag renders only the element whose data changed (61 renders).

Three sources, all fixed:

1. **PlayProvider** stored live cursor positions in unmemoized context state, so every presence message produced a new context value and re-rendered every context consumer (every `CanPlayElement`). The context value is now memoized; the presence map keeps a stable identity and is mutated in place. Reactive consumers use the `useCursorPresences` hook, which subscribes to the cursor client directly and is unchanged.
2. **`usePlayerIdentity`** set a fresh state object on every `users.onChange` tick (which fires per presence update) even when the identity was unchanged. It now bails when color/pid/name are identical.
3. **The fridge page**: `WordControls` — the component that renders the entire word list — called `useCursorPresences()` only to display an online count, re-rendering all words per tick. The count is now an isolated `<OnlineCount />` component.

## Rationale

Found while perf-testing the v2 protocol branch against the full production fridge room seeded locally: the storm exists identically on v1 and was masked only by small rooms. Production pays this cost today on any large room with cursors enabled.

## Measurements (fridge room, 2,906 mounted words, 1s drag, ~120 presence ticks)

| | before | after |
|---|---|---|
| PlayProvider renders | ~120 | 1 |
| CanPlayElement renders | ~340,000 | 61 (dragged word only) |
| word-list re-renders | ~120 | 0 |

No visual/UI change (perf only), so no screenshots; the render-count instrumentation methodology is reproducible via the latency probe on the v2 branch (`feat/v2-op-protocol`, `e2e/v2/latency-probe.ts`).

## Notes

- Changeset included (`@playhtml/react` patch).
- `PlayContextInfo` shape is unchanged; `context.cursorPresences` still exists but its Map identity is now stable (nothing in this repo consumed it reactively — the documented path is `useCursorPresences`).
- Tests: `bun run test:react` green (52 + 2), `tsc --noEmit` clean, website vite build passes.

### DIFF
--- a/.changeset/react-presence-render-storm.md
+++ b/.changeset/react-presence-render-storm.md
@@ -1,0 +1,5 @@
+---
+"@playhtml/react": patch
+---
+
+Stop cursor-presence updates from re-rendering every playhtml component on the page. PlayProvider no longer stores live cursor positions in context state (the context value is now memoized and the presence map keeps a stable identity; reactive consumers use the useCursorPresences hook, which is unchanged), and usePlayerIdentity only re-renders its consumers when the identity actually changes rather than on every presence tick. On a room with ~3,000 elements this reduced renders during a one-second drag from ~340,000 to just the elements whose data changed.
--- a/packages/react/src/PlayProvider.tsx
+++ b/packages/react/src/PlayProvider.tsx
@@ -247,19 +247,33 @@ export function PlayProvider({
     name: undefined,
   });
 
-  const [cursorPresences, setCursorPresences] = useState<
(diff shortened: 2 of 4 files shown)


spencerc99:
## Code review

PR #397, opened 2026-08-18
extension: make the opt-in new tab redirect work on Firefox, honest Safari fallback

spencerc99:
## Summary

Pre-release hardening of the opt-in new tab takeover (#391), found by empirical testing in real browsers:

1. **Grandfather/redirect race (all browsers).** The migration writes `newtab_takeover_enabled` during worker startup, racing the redirect module's one-time cached read and its `onChanged` listener registration — captured live: the cache stayed `false` after the migration wrote `true`. Existing users' first post-update tabs would not have redirected. The preference is now read per new tab (once per tab; no hot-path cost that matters).
2. **Firefox event non-determinism.** In one session, a new tab reported `about:newtab` at `onCreated` in one trial and `about:blank` (URL arriving only via `onUpdated`) in the next. Both paths handled; `about:home` added; the first real URL on a blank tab decides — a navigation to anything else leaves the tab alone. Never redirects twice.
3. **Safari fallback.** Safari cannot replace the new tab, so the setup step keeps the preview but swaps the checkbox for a bookmark/pin suggestion (writes `false`); same on the history page. Non-Safari behavior unchanged.

## Evidence

- Firefox (web-ext, real browser): 5/5 new tabs land on `walking-record.html`
- Chromium (playwright + CDP-created `chrome://newtab/` tabs): 3/3, no regression
- Extension console output had to be captured via a local log collector — web-ext does not relay it to stdout

## Testing

784 passing (123 files); takeover unit tests rewritten to encode observed Firefox behavior; two vacuous Safari setup tests (asserting non-Safari UI under a Safari mock) moved and fixed. `check:extension` clean; Chrome/Firefox/Safari builds succeed.

No changeset (extension-only). No new PENDING bullet: this fixes behavior that ships in the same release as the existing bullets.


spencerc99:
## Code review

- [P2] Add the required screenshots for the new Safari-only UI — [`extension/src/components/SetupPage.tsx:542-545`](https://github.com/spencerc99/playhtml/blob/65f417b216fcd6230495bf98fc8805700f3074c0/extension/src/components/SetupPage.tsx#L542-L545) and [`extension/src/entrypoints/walking-record/walking-record.tsx:382-387`](https://github.com/spencerc99/playhtml/blob/65f417b216fcd6230495bf98fc8805700f3074c0/extension/src/entrypoints/walking-record/walking-record.tsx#L382-L387). These changed branches introduce visible Safari-specific replacement copy on two extension surfaces, but the PR has no screenshots of either surface. The root `AGENTS.md` explicitly requires every user-facing change to be verified on the real affected surface and accompanied by screenshots. Attach screenshots from the real Safari extension showing the setup new-tab step and the walking-record bookmark/pin notice.

PR #404, opened 2026-08-20
playhtml: reduce repeated element work

spencerc99:
## Summary

- Skip PlayHTML handlers already bound to the same DOM node during repeated document scans while preserving explicit reinitialization for room changes and server resets.
- Deliver local element-awareness changes only to the affected element and fingerprint the full awareness snapshot once per coalesced publish.
- Document stable element IDs and add a patch changeset for the published runtime behavior.

## Why

React route changes repeatedly scan registered elements. The previous scan path set up every matching element again, causing work and renders proportional to the total registered element count even when the DOM nodes had not changed.

Local element-awareness writes also rebuilt and fingerprinted the complete awareness map for every update. This made per-element writes scale with the total element count.

## Measurements

All browser figures are ranges from five Chrome runs on the same machine, Bun version, lockfile, build, and benchmark paths.

| Public path | Before | After |
| --- | ---: | ---: |
| React route scan, 1,000 elements | 29.6-36.0 ms, 1,000 renders | 0.2-0.4 ms, 0 renders |
| Element-awareness initialization, 1,000 elements | 1,285.5-1,328.6 ms | 901.6-912.7 ms |
| 50 local awareness writes, 1,000 elements | 37.3-39.6 ms | 0.1-0.2 ms |

The public ESM path grows by 1,677 raw bytes, 362 gzip bytes, and 374 Brotli bytes using the same build and compression commands.

## Reproduction

- Create 1,000 public React capability elements, initialize PlayHTML, and trigger a route search without replacing the nodes. Record elapsed time and component render callbacks.
- Create 1,000 public `can-play` elements with element awareness, initialize them, then perform 50 local awareness writes. Record initialization and write time.
- Repeat each scenario in five fresh browser runs.

## Verification

- `bun run build-packages`
- PlayHTML tests: 517 passed
- React tests: 54 passed
- Docs tests: 52 passed
- Docs static build
- `bun run check:packages`
- Five-run Chrome route, awareness, and lifecycle measurements

## Scope notes

The lifecycle test also confirmed that unique element IDs in one long-lived room retain CRDT history after live records are deleted. Reclaiming that history requires a separate document-lifecycle decision and is not part of this fix.

The public docs describe repeated scans and stable IDs. The starter templates are unchanged because this does not change API signatures or usage. There is no visual UI change, so screenshots are not applicable.

### DIFF
--- a/.changeset/quiet-elements-scale.md
+++ b/.changeset/quiet-elements-scale.md
@@ -1,0 +1,5 @@
+---
+"playhtml": patch
+---
+
+Avoid reinitializing bound elements during repeated page scans and limit local awareness updates to the element that changed, reducing navigation and presence overhead on pages with many PlayHTML elements.
--- a/apps/docs/src/content/docs/advanced/dynamic-elements.mdx
+++ b/apps/docs/src/content/docs/advanced/dynamic-elements.mdx
@@ -52,6 +52,8 @@ el.remove();
 playhtml.deleteElementData("can-move", "note-42");
 ```
 
+Reuse the same stable `id` when a logical item can return. A fresh `id` creates a separate persisted item on every mount. Call `deleteElementData` when the item is permanently removed.
+
 See [Development & data cleanup](/docs/advanced/development/) for inspecting and resetting stored state from the devtools panel.
 
   </TabItem>
--- a/apps/docs/src/content/docs/reference/playhtml-client.md
+++ b/apps/docs/src/content/docs/reference/playhtml-client.md
@@ -152,7 +152,7 @@ playhtml.setupPlayElement(card);
 
 **Signature:** `setupPlayElements(): void`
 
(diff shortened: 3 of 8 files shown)


spencerc99:
## Code review

- **[P2] Re-key every local element when the player identity changes.** In `packages/playhtml/src/element-awareness.ts:324-336`, `emitLocalAwareness()` filters the previous identity through the shared `localStableId`, then updates that field after processing only one element. With two local elements keyed under identity A, changing to identity B and updating both before a server echo makes the first update replace A with B, but the second now filters only B and preserves its stale A entry before adding B. That element's awareness callback therefore receives the same local user under both identities. Identity replacement is reachable through `users.adoptIdentity()` and the extension listener at `packages/playhtml/src/index.ts:1349-1364`; the rejoin at `packages/playhtml/src/index.ts:676-692` only sends to the transport and does not synchronously rebuild the peer map. Detect the identity change before targeted delivery and rebuild or re-key the complete local awareness map once, then update `localStableId`; add a regression with two elements updated after an identity change and before any transport response.

PR #406, opened 2026-08-21
common: avoid redundant shared-data writes

spencerc99:
## Summary

- skip writes when incoming shared-data primitives have not changed
- preserve equal arrays instead of replacing them in SyncedStore proxies
- compare nested plain objects without depending on property insertion order

## Why

Room bridges send complete shared-element subtrees. `deepReplaceIntoProxy` previously rewrote every primitive and array in an incoming subtree, even when only property order differed or one nested value changed. Each rewrite appended Yjs history.

The production showcase room amplified semantically unchanged `student-projects` updates into rapid document growth, repeated autosaves, compactions, and Durable Object memory resets.

## Verification

- `bun run test:common`: 18 tests passed
- `bun run --cwd packages/common build`: passed
- `tsc -p partykit`: passed
- production Worker dry-run bundle: passed
- deployed to staging as `a618c71f-9d42-4c0b-831a-edf85d91cfcf`
- deployed to production as `a34c9773-52bc-4cde-ac91-e481799c0150`
- restored the production showcase document with all 27 projects preserved
- observed zero document growth and no semantic change across a 45-second bridge interval
- observed zero targeted failure signatures and zero runtime errors after recovery on the deployed version

## Documentation

No public documentation or starter-template change is needed. This restores the intended bridge behavior without changing the shared-data API.

### DIFF
--- a/.changeset/quiet-bridge-updates.md
+++ b/.changeset/quiet-bridge-updates.md
@@ -1,0 +1,5 @@
+---
+"@playhtml/common": patch
+---
+
+Avoid redundant shared-data writes when incoming object and array values are unchanged, preventing unnecessary collaborative document history growth.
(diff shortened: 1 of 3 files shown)


spencerc99:
## Code review

PR #416, opened 2026-08-26
Protect room persistence recovery from stale writes

spencerc99:
## Summary

- serialize document saves, restores, compaction, transient recovery, and bridge applies behind one room write boundary
- keep shared-data writes paused until authoritative persistence recovery completes, then reset connected clients onto the recovered snapshot
- authenticate cross-room bridge requests and preserve queued bridge updates across maintenance
- add regression coverage and real bridge, transient, stale-source, destroy/reconnect, and compaction smoke tests

## Why

A stale or concurrent write could race an authoritative restore or compaction and overwrite recovered room data. The room now admits persistence-changing operations through one state machine and durable recovery path, including restart and alarm recovery cases.

## Verification

Local:
- bun test partykit/__tests__/*.test.ts: 274 pass, 0 fail, 1397 expectations
- bun run check:partykit
- Prettier and git diff --check

Real staging matrix:
- authenticated bridge hydration and bidirectional updates; missing token 401, wrong token 403
- server request/rate limits and sustained normal traffic
- 8-client keyed-write soak: 1,600 upserts, 8 converged records, bounded Yjs history
- 20-client presence soak at 60 Hz for 20 seconds
- live bridge hibernation and provider destroy/reconnect after 90 seconds
- empty-room compaction after the full 5-minute alarm, stale-client reset rejection, fresh reconnect, and second-window no-op
- stale live compaction source versus newer persisted snapshot after the full 5-minute alarm
- local Supabase-unavailable transient mode: awareness remains live; shared-data and admin writes remain blocked
- connected-room emergency compaction with a temporary staging threshold: active client reset/disconnect, state preserved, fresh reconnect
- restored standard staging deployment and reran the authenticated bridge integration smoke

Final standard staging version: 20322937-9f07-49cf-92cc-edbf285c1dd3. The only active staging quarantine is the pre-existing drill fixture drill-oom-mega; no matrix room was quarantined. Production was not changed.

No changeset, public docs, starter-template update, or screenshot is needed because this changes the internal PartyKit persistence runtime and smoke tooling, not a published package API or user-facing UI.

### DIFF
--- a/partykit/bridgeLeasePolicy.ts
+++ b/partykit/bridgeLeasePolicy.ts
@@ -1,6 +1,8 @@
 // ABOUTME: Merges and renews shared-reference leases for active bridge consumers.
 // ABOUTME: Preserves source metadata while extending only requested relationships.
-import type { SharedRefEntry } from "./const";
+import { DEFAULT_SUBSCRIBER_LEASE_MS, type SharedRefEntry } from "./const";
+
+const SHARED_REFERENCE_RENEWAL_MS = DEFAULT_SUBSCRIBER_LEASE_MS / 2;
 
 type SharedReferenceRequest = {
   sourceRoomId: string;
@@ -28,19 +30,37 @@ export function mergeSharedReferenceLeases({
   const requestedBySource = new Map(
     requested.map((entry) => [entry.sourceRoomId, entry.elementIds] as const)
   );
+  let changed = false;
+  const now = Date.parse(nowIso);
   const entries = existing.map((entry) => {
     const requestedIds = requestedBySource.get(entry.sourceRoomId);
     if (requestedIds === undefined) return { ...entry };
 
     requestedBySource.delete(entry.sourceRoomId);
+    const elementIds = Array.from(
+      new Set([...entry.elementIds, ...requestedIds])
+    );
+    const lastSeen = Date.parse(entry.lastSeen ?? "");
+    const membershipChanged = elementIds.length !== entry.elementIds.length;
+    const leaseNeedsRenewal =
+      !Number.isFinite(lastSeen) ||
+      !Number.isFinite(now) ||
+      now - lastSeen >= SHARED_REFERENCE_RENEWAL_MS;
+
+    if (!membershipChanged && !leaseNeedsRenewal) {
+      return { ...entry };
+    }
+
+    changed = true;
     return {
       ...entry,
-      elementIds: Array.from(new Set([...entry.elementIds, ...requestedIds])),
+      elementIds,
       lastSeen: nowIso,
     };
   });
 
   for (const [sourceRoomId, elementIds] of requestedBySource) {
+    changed = true;
     entries.push({
       sourceRoomId,
       elementIds: Array.from(new Set(elementIds)),
@@ -48,5 +68,5 @@ export function mergeSharedReferenceLeases({
     });
   }
 
-  return { entries, changed: true };
+  return { entries, changed };
 }
--- a/partykit/bridgeLeasePolicy.ts
+++ b/partykit/bridgeLeasePolicy.ts
@@ -1,0 +1,52 @@
+// ABOUTME: Merges and renews shared-reference leases for active bridge consumers.
+// ABOUTME: Preserves source metadata while extending only requested relationships.
+import type { SharedRefEntry } from "./const";
+
+type SharedReferenceRequest = {
+  sourceRoomId: string;
+  elementIds: string[];
+};
+
+type MergeSharedReferenceLeasesOptions = {
+  existing: readonly SharedRefEntry[];
+  requested: readonly SharedReferenceRequest[];
+  nowIso: string;
+};
+
+export function mergeSharedReferenceLeases({
+  existing,
+  requested,
+  nowIso,
+}: MergeSharedReferenceLeasesOptions): {
+  entries: SharedRefEntry[];
+  changed: boolean;
+} {
+  if (requested.length === 0) {
+    return { entries: [...existing], changed: false };
+  }
+
+  const requestedBySource = new Map(
+    requested.map((entry) => [entry.sourceRoomId, entry.elementIds] as const)
+  );
+  const entries = existing.map((entry) => {
+    const requestedIds = requestedBySource.get(entry.sourceRoomId);
+    if (requestedIds === undefined) return { ...entry };
+
+    requestedBySource.delete(entry.sourceRoomId);
+    return {
+      ...entry,
+      elementIds: Array.from(new Set([...entry.elementIds, ...requestedIds])),
+      lastSeen: nowIso,
+    };
+  });
+
+  for (const [sourceRoomId, elementIds] of requestedBySource) {
+    entries.push({
+      sourceRoomId,
+      elementIds: Array.from(new Set(elementIds)),
+      lastSeen: nowIso,
+    });
+  }
+
+  return { entries, changed: true };
+}
--- a/partykit/party.ts
+++ b/partykit/party.ts
@@ -159,6 +159,8 @@ type DocumentWriteState =
 
 type PendingDocumentReset = {
   resetEpoch: number;
+  attempt?: number;
+  retryAt?: number;
 };
 
 type DocumentSaveRetry = {
@@ -888,23 +890,81 @@ export class PartyServer extends YServer {
       "resetEpoch" in value &&
       typeof value.resetEpoch === "number"
     ) {
-      return { resetEpoch: value.resetEpoch };
+      return {
+        resetEpoch: value.resetEpoch,
+        attempt:
+          "attempt" in value && typeof value.attempt === "number"
+            ? value.attempt
+            : undefined,
+        retryAt:
+          "retryAt" in value && typeof value.retryAt === "number"
+            ? value.retryAt
+            : undefined
+      };
     }
     return null;
   }
 
-  private async reconcilePendingDocumentReset(): Promise<void> {
+  private async schedulePendingDocumentResetRetry(
+    resetEpoch: number
+  ): Promise<void> {
+    try {
+      const previous = await this.getPendingDocumentReset();
+      const attempt = (previous?.attempt ?? 0) + 1;
+      const delay = Math.min(
+        DEFAULT_DOCUMENT_SAVE_RETRY_MS * 2 ** (attempt - 1),
+        DEFAULT_DOCUMENT_SAVE_RETRY_MAX_MS
+      );
+      await this.ctx.storage.put(STORAGE_KEYS.pendingDocumentReset, {
+        resetEpoch,
+        attempt,
+        retryAt: Date.now() + delay
+      } satisfies PendingDocumentReset);
+      await this.scheduleNextAlarm();
+    } catch (error) {
+      console.error(
+        `[PartyServer] Pending document reset retry scheduling failed for room=${this.name}:`,
+        error
+      );
+    }
+  }
+
+  private async reconcilePendingDocumentReset(): Promise<boolean> {
     const pending = await this.getPendingDocumentReset();
-    if (pending === null) return;
+    if (pending === null) return true;
 
     const documentResetEpoch = getDocResetEpoch(this.document);
-    if (documentResetEpoch === pending.resetEpoch) {
+    if (documentResetEpoch !== pending.resetEpoch) {
+      this.documentWriteState = {
+        kind: "recovery-required",
+        resetEpoch: pending.resetEpoch
+      };
+      await this.schedulePendingDocumentResetRetry(pending.resetEpoch);
+      return false;
+    }
+
+    try {
       await this.setResetEpoch(pending.resetEpoch);
+      await this.ctx.storage.delete(STORAGE_KEYS.pendingDocumentReset);
+      if (this.getDocumentWriteState().kind === "recovery-required") {
+        this.documentWriteState = { kind: "idle" };
+      }
       console.warn(
         `[PartyServer] Recovered interrupted document reset for room=${this.name}: resetEpoch=${pending.resetEpoch}`
       );
+      return true;
+    } catch (error) {
+      this.documentWriteState = {
+        kind: "recovery-required",
+        resetEpoch: pending.resetEpoch
+      };
+      console.error(
+        `[PartyServer] Pending document reset reconciliation failed for room=${this.name}:`,
+        error
+      );
+      await this.schedulePendingDocumentResetRetry(pending.resetEpoch);
+      return false;
     }
-    await this.ctx.storage.delete(STORAGE_KEYS.pendingDocumentReset);
   }
 
   private async getPersistedDocumentBase64(): Promise<string | null> {
@@ -986,7 +1046,6 @@ export class PartyServer extends YServer {
     afterReplace
   }: CommitCompactedDocumentOptions): Promise<boolean> {
     this.documentWriteState = { kind: "compaction" };
-    let liveDocumentReplaced = false;
     let databaseCommitted = false;
 
     try {
@@ -1058,15 +1117,26 @@ export class PartyServer extends YServer {
         kind: "recovery-required",
         resetEpoch: compactedDocument.resetEpoch
       };
-      await this.setResetEpochAfterDocumentCommit(
+      const resetEpochCommitted = await this.setResetEpochAfterDocumentCommit(
         compactedDocument.resetEpoch
       );
 
       this.compactionAutosaveSnapshot = compactedDocument.base64;
       replaceDocFromSnapshot(this.document, compactedDocument.base64);
-      liveDocumentReplaced = true;
-      await this.ctx.storage.delete(STORAGE_KEYS.pendingDocumentReset);
-      this.documentWriteState = { kind: "idle" };
+      if (resetEpochCommitted) {
+        this.documentWriteState = { kind: "idle" };
+        try {
+          await this.ctx.storage.delete(STORAGE_KEYS.pendingDocumentReset);
+        } catch (error) {
+          console.error(
+            `[PartyServer] Compaction marker cleanup failed for room=${this.name}:`,
+            error
+          );
+          await this.schedulePendingDocumentResetRetry(
+            compactedDocument.resetEpoch
+          );
+        }
+      }
       try {
         await afterReplace?.();
       } catch (error) {
@@ -1084,10 +1154,7 @@ export class PartyServer extends YServer {
       }
       throw error;
     } finally {
-      if (
-        liveDocumentReplaced ||
-        this.getDocumentWriteState().kind === "compaction"
-      ) {
+      if (this.getDocumentWriteState().kind === "compaction") {
         this.documentWriteState = { kind: "idle" };
       }
     }
@@ -1113,12 +1180,14 @@ export class PartyServer extends YServer {
     }
   }
 
-  private async setResetEpochAfterDocumentCommit(epoch: number): Promise<void> {
+  private async setResetEpochAfterDocumentCommit(
+    epoch: number
+  ): Promise<boolean> {
     let lastError: unknown;
     for (let attempt = 1; attempt <= RESET_EPOCH_WRITE_ATTEMPTS; attempt += 1) {
       try {
         await this.setResetEpoch(epoch);
-        return;
+        return true;
       } catch (error) {
         lastError = error;
         if (attempt < RESET_EPOCH_WRITE_ATTEMPTS) {
@@ -1129,7 +1198,12 @@ export class PartyServer extends YServer {
         }
       }
     }
-    throw lastError;
+    console.error(
+      `[PartyServer] Reset epoch remains pending for room=${this.name}:`,
+      lastError
+    );
+    await this.schedulePendingDocumentResetRetry(epoch);
+    return false;
   }
 
   private setConnectionAcceptedResetEpoch(
@@ -1352,8 +1426,11 @@ export class PartyServer extends YServer {
       console.log(`[Restore Snapshot] Successfully saved snapshot to database`);
 
       // Set reset epoch for client detection
-      await this.setResetEpochAfterDocumentCommit(resetEpoch);
-      console.log(`[Restore Snapshot] Set resetEpoch: ${resetEpoch}`);
+      const resetEpochCommitted =
+        await this.setResetEpochAfterDocumentCommit(resetEpoch);
+      if (resetEpochCommitted) {
+        console.log(`[Restore Snapshot] Set resetEpoch: ${resetEpoch}`);
+      }
 
       // Broadcast a "room-reset" message to all connected clients
       this.broadcastCustomMessage(this.getRoomResetMessage(resetEpoch));
@@ -1378,8 +1455,18 @@ export class PartyServer extends YServer {
       replaceDocFromSnapshot(liveYDoc, updatedBase64);
       setDocResetEpoch(liveYDoc, resetEpoch);
       this.markDocumentHydrated();
-      await this.ctx.storage.delete(STORAGE_KEYS.pendingDocumentReset);
-      resetCompleted = true;
+      if (resetEpochCommitted) {
+        try {
+          await this.ctx.storage.delete(STORAGE_KEYS.pendingDocumentReset);
+          resetCompleted = true;
+        } catch (error) {
+          console.error(
+            `[Restore Snapshot] Marker cleanup failed for room=${roomId}:`,
+            error
+          );
+          await this.schedulePendingDocumentResetRetry(resetEpoch);
+        }
+      }
       console.log(`[Restore Snapshot] Successfully reloaded live server`);
 
       console.log(
@@ -1498,16 +1585,26 @@ export class PartyServer extends YServer {
     const documentSaveRetry = await this.getDocumentSaveRetry();
     const persistenceRecoveryAlarm =
       await this.circuitBreaker.getPersistenceRecoveryRetryAfter();
-    const nextAlarm = [
-      maintenanceAlarm,
-      documentSaveRetry?.retryAt ?? null,
-      persistenceRecoveryAlarm,
-    ].reduce<
-      number | null
-    >((earliest, candidate) => {
-      if (candidate === null) return earliest;
-      return earliest === null ? candidate : Math.min(earliest, candidate);
-    }, null);
+    const pendingDocumentReset = await this.getPendingDocumentReset();
+    const pendingDocumentResetAlarm =
+      pendingDocumentReset === null
+        ? null
+        : (pendingDocumentReset.retryAt ?? now);
+    const nextAlarm =
+      [
+        maintenanceAlarm,
+        documentSaveRetry?.retryAt ?? null,
+        persistenceRecoveryAlarm,
+        pendingDocumentResetAlarm,
+      ].reduce<number | null>(
+        (earliest, candidate) =>
+          candidate === null
+            ? earliest
+            : earliest === null
+              ? candidate
+              : Math.min(earliest, candidate),
+        null
+      );
 
     if (nextAlarm === null) {
       await this.ctx.storage.deleteAlarm?.();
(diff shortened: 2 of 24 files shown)


spencerc99:
## Code review

- **P2 — A new failure-path test leaves its expected error uncaptured.** [`partykit/__tests__/quarantineLoad.test.ts:2405`](https://github.com/spencerc99/playhtml/blob/d3e188f1132159eba86740c5ae7c1a665f25daf8/partykit/__tests__/quarantineLoad.test.ts#L2405-L2421) deliberately throws `simulated isolate crash before database commit` and asserts only the rejection. The exact-head PartyKit suite passes but prints that error and stack trace to stderr, violating the repository requirement that test output be pristine and expected errors be captured and validated. Spy on the expected `console.error` or `console.warn`, assert the intended diagnostic, and restore it after the test; apply the same treatment to any other added failure scenarios that intentionally log.

partykit/party.ts
```diff
@@ -299,8 +346,17 @@ export class PartyServer extends YServer {
...
-    if (await this.circuitBreaker.shouldDeferLoad()) {
+    const persistenceRecoveryPending =
+      (await this.ctx.storage.get(
+        STORAGE_KEYS.persistenceRecoveryPending
+      )) === true;
+
+    // Persistence recovery has its own durable deadline and must reach onLoad
+    // before the ordinary load-failure ledger can defer startup.
+    if (
+      !persistenceRecoveryPending &&
```
spencerc99: **[severe] Recovery hydration has no vanish protection: an OOM during recovery becomes an unbounded crash loop with quarantine suppressed.**

While `persistenceRecoveryPending` is set, three protections are disabled at once:

1. `onStart` skips `shouldDeferLoad()` entirely (this line), so there is no load backoff between attempts.
2. `shouldDeferLoad` / `getLoadDeferredResponse` suppress `shouldQuarantineForFailures` when `recoveryPending` is true (roomCircuitBreaker.ts:392-397, 614-621), so the room can never quarantine no matter how many load attempts accumulate.
3. `persistenceRecoveryRetryAfter` is only advanced inside catch blocks (`schedulePersistenceRecovery` in `enterTransientPersistenceMode` and `recoverTransientPersistenceIfDue`). A **vanish** (isolate killed mid-hydration) observes no exception, so the deadline stays in the past.

Failure scenario: a room with a large persisted document hits a transient Supabase outage → `persistenceRecoveryPending` is set. The recovery path (`loadDocument(true)` → `restoreFromSnapshotNow`) holds roughly two Y.Doc copies plus two base64 strings — more memory than a normal load — and OOM-kills the isolate. On every subsequent wake (connect, request, alarm) the retry deadline is already due, backoff is skipped, quarantine is suppressed, and the room immediately re-attempts the same OOM-ing restore. `loadAttempts` grows forever but nothing acts on it. The pre-PR restart path would have backed off and quarantined after the threshold — that vanish accounting (the whole point of `beginRiskyOperation("load")`'s durable increment) is bypassed exactly for the doomed case. Only a manual operator quarantine escapes the loop.

A fix: advance `persistenceRecoveryRetryAfter` durably *before* attempting the recovery restore (begin/complete pattern, like `beginRiskyOperation`), and let recovery-pending rooms still quarantine after N vanished attempts.

---

partykit/party.ts
```diff
@@ -2819,9 +3377,32 @@ export class PartyServer extends YServer {
...
+      ) {
+        await this.runDocumentWrite(() =>
+          this.reconcilePendingDocumentReset()
+        );
+      }
+
+      const documentSaveRetry = await this.getDocumentSaveRetry();
+      if (
+        documentSaveRetry !== null &&
+        documentSaveRetry.retryAt <= Date.now()
```
spencerc99, line 3070: **[correctness] Hot alarm loop: a due `documentSaveRetry` plus a silently-skipped autosave re-arms the alarm at a past timestamp forever.**

When the retry is due and `roomState() === "ready"`, this branch calls `persistLiveDocument`. But `persistLiveDocumentNow` has skip paths that `return false` without clearing or rescheduling the retry record — notably the `getAutosaveResetEpochDecision` "skip" case (doc reset epoch missing or older than the stored epoch). The sequence then is:

1. Alarm fires, retry record has `retryAt` in the past.
2. Autosave skips silently; retry record untouched.
3. `onAlarm`'s `finally` calls `scheduleNextAlarm()`, which includes `documentSaveRetry.retryAt` (past) as a candidate; inside the alarm handler `getAlarm()` returns null, so `shouldSetAlarm` passes and `setAlarm(past)` is armed.
4. The alarm fires again immediately → goto 2.

The epoch-stale condition is persistent (it doesn't resolve on its own), so this is a tight, indefinite alarm loop burning Durable Object invocations and log volume. The `consumeCompactionAutosave` early-return has the same shape but self-resolves after one extra cycle.

Fix: any `return false`/skip path in `persistLiveDocumentNow` that is not a genuine transient failure should either clear the retry record or reschedule it with backoff (same treatment the `state !== "ready"` branch gets here in `onAlarm`).

---

partykit/party.ts
```diff
@@ -2083,19 +2595,26 @@ export class PartyServer extends YServer {
...
         `[PartyServer] SUPABASE AUTOSAVE FAILED for room ${this.name}:`,
         error
       );
+      await this.scheduleDocumentSaveRetry();
       return false;
     }
+    if ((this.documentGeneration ?? 0) === saveGeneration) {
+      await this.clearDocumentSaveRetry();
+    } else {
+      await this.scheduleDocumentSaveRetry();
```
spencerc99: **[correctness/efficiency] Successful saves escalate the failure-backoff counter.**

This branch runs after a save that *succeeded* — the doc merely changed during the upsert, which is routine in any active room. `scheduleDocumentSaveRetry()` increments the same `attempt` counter used for real failures, and the counter is only cleared when a save completes with an unchanged generation. In a continuously-busy room:

- the attempt counter ratchets up on every autosave and never resets, so follow-up saves drift to the 60s max delay even though every save is succeeding;
- when a genuine Supabase failure later occurs, its first retry starts at max backoff (60s) instead of `DEFAULT_DOCUMENT_SAVE_RETRY_MS` (1s);
- every autosave in an active room now pays an extra storage read+write and alarm arming, duplicating work y-partyserver's own debounced `onSave` already does for new updates.

Suggest treating success-with-newer-data as a fresh follow-up (reset `attempt` to 0 / schedule at base delay), keeping exponential backoff only for actual save failures.

---

partykit/party.ts
```diff
@@ -753,6 +989,42 @@ export class PartyServer extends YServer {
...
     options: SaveDocumentOptions = {}
   ): Promise<void> {
+    await this.runDocumentWrite(async () => {
+      this.assertDocumentSaveAllowed(options);
+      if (
+        (options.operation ?? "shared-data") === "shared-data" &&
+        documentBase64 !== encodeDocToBase64(this.document)
+      ) {
+        throw new Error(
+          "Cannot persist stale document: live room state changed before the save acquired the write queue"
```
spencerc99: **[correctness] Admin force-save-live now 500s spuriously in active rooms.**

`handleAdminForceSaveLive` (admin.ts:452-453) encodes the live doc *before* calling `saveDocumentBase64`, which then waits for the write queue and re-encodes for this staleness comparison. Any client write landing between the handler's encode and the queue slot (e.g. while a queued autosave's Supabase upsert is in flight) makes the two encodings differ, and the save throws → the operator gets a 500. The escape-hatch endpoint becomes unreliable exactly in the busy rooms where an operator is most likely to need it, and retrying has the same race.

Since the intent of force-save-live is "persist whatever the live doc holds now," the fix is to capture the encoding *inside* the queued closure (accept a thunk, or have the admin route call `persistLiveDocument({ allowCompaction: false })`) rather than rejecting the operator's snapshot as stale.

---

partykit/bridgeLeasePolicy.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  for (const [sourceRoomId, elementIds] of requestedBySource) {
+    entries.push({
+      sourceRoomId,
+      elementIds: Array.from(new Set(elementIds)),
+      lastSeen: nowIso,
+    });
+  }
+
+  return { entries, changed: true };
```
spencerc99: **[efficiency/behavior] `changed: true` on every non-empty request re-triggers full subscribe+hydrate fan-out.**

The old merge returned `changed: false` when the requested references already existed; now any non-empty request reports `changed: true` (correct for lease renewal, but the flag no longer means "entries changed"). The consumer at party.ts:1896-1899 uses it as a hydration trigger:

```ts
const { entries, changed } = await this.mergeAndStoreSharedRefs([...]);
if (changed) await this.subscribeAndHydrate(entries);
```

`handleRegisterSharedReference` runs once per shared-reference message, and `subscribeAndHydrate(entries)` subscribes to **every** known source room, not just the one in this message. So a returning visitor to a page with K shared references now causes K register messages × K sources = O(K²) cross-room subscribe HTTP calls plus repeated subtree hydration per page load — previously these were no-ops for already-known references (and the onConnect path at party.ts:2043-2046 already does one full subscribe+hydrate per connection).

Suggest returning two signals (`entriesChanged` vs `leaseRenewed`) — or having the caller hydrate only newly-added sourceRoomIds — so lease renewal doesn't amplify bridge traffic.

---

partykit/party.ts
```diff
@@ -188,26 +223,37 @@ function readPositiveNumberEnv(name: string, fallback: number): number {
...
   private compactionAutosaveSnapshot: string | null = null;
   private cachedResetEpoch: number | null | undefined;
   private lastKnownDocumentBytes = 0;
   private hasWarnedDocumentSize = false;
 
+  public get isSkippingSave(): boolean {
+    return this.getDocumentWriteState().kind !== "idle";
+  }
+
+  public set isSkippingSave(paused: boolean) {
```
spencerc99: **[hazard] `isSkippingSave` setter silently bypasses the write-state machine.**

`isSkippingSave = false` forces `{ kind: "idle" }` unconditionally. If the current state is `recovery-required` (a committed reset whose durable marker/epoch cleanup is still pending) or an in-flight `compaction`/`reset`, this re-enables autosave against exactly the window the new state machine exists to protect — the next autosave could overwrite the recovered snapshot with a doc whose reset bookkeeping is incomplete.

Production code no longer uses the setter (only tests do), so today it's latent — but it's a `public` property on the server class with the old, innocuous name. Any future caller (or a test asserting through it) reintroduces the stale-write race. Suggest removing it (update the tests to drive real paths), or at minimum making `= false` a no-op unless the state is `maintenance`.

---

partykit/party.ts
```diff
@@ -1781,8 +2254,8 @@ export class PartyServer extends YServer {
           );
         },
       }
-    ).catch((error) => {
-      this.enterTransientPersistenceMode(error);
+    ).catch(async (error) => {
+      await this.enterTransientPersistenceMode(error);
```
spencerc99: **[robustness] `enterTransientPersistenceMode` can now reject, turning graceful degradation into a failed load.**

Previously this was synchronous and could not fail; now it awaits `ctx.storage.put` + `schedulePersistenceRecovery` + `scheduleNextAlarm` inside the Supabase-load `.catch`. If any of those storage writes throws, the rejection propagates out of the catch handler, `result` rejects, and `loadDocument` throws — so the exact situation this path exists for (degrade to awareness-only transient mode when persistence is down) instead fails room startup outright, and in `onStart` context the exception escapes `onLoad`. Worth wrapping the durable bookkeeping in its own try/catch so the in-memory transient mode (already set on the first lines) always survives even if the durable recovery scheduling could not be recorded.

---

spencerc99 (commented):
Single-pass high-effort review (the multi-agent fan-out + subagent verify pass could not run in this environment — all 8 angles were worked through inline by one reviewer, so treat confidence accordingly). 7 findings, ranked most-severe first; details in the inline comments.

1. Recovery hydration has no vanish/OOM protection: with `persistenceRecoveryPending` set, load backoff is skipped, quarantine is suppressed, and the recovery deadline only advances on observed exceptions — an OOM during the recovery restore becomes an unbounded crash loop (party.ts:357).
2. Hot alarm loop: a due `documentSaveRetry` combined with a silent autosave skip (stale doc reset epoch) re-arms the alarm at a past timestamp indefinitely (party.ts:3397).
3. Successful saves escalate the save-retry failure counter, pinning follow-up saves at max backoff in busy rooms (party.ts:2616).
4. Admin force-save-live can spuriously 500: the doc is encoded before the write queue, so any concurrent client write trips the new staleness guard (party.ts:999 / admin.ts:452).
5. `mergeSharedReferenceLeases` reports `changed: true` for no-op re-registrations, so every register-shared-reference message re-runs subscribe+hydrate against all sources — O(K²) bridge calls per page load (bridgeLeasePolicy.ts:51).
6. The retained `isSkippingSave` setter bypasses the new write-state machine and can clear `recovery-required` (party.ts:243).
7. `enterTransientPersistenceMode` is now awaited fallible storage work inside the load `.catch`; a storage error turns graceful transient degradation into a failed load (party.ts:2258).

The core design — one write queue, durable pending-reset marker, epoch-after-commit ordering, authoritative-snapshot recovery, bridge auth + relationship enforcement — checked out against the failure orderings I traced (isolate death between marker/DB/epoch writes, alarm-during-restore interleavings, queue re-entrancy/deadlock, rollback paths). No deadlocks or lost-reset windows found beyond the items above.

---

spencerc99:
## Simplification review (follow-up to the 7 inline findings)

This PR's core idea — one serialized write boundary (`runDocumentWrite`) for all document persistence — is sound and fixes real races. But the PR then layers four independent retry/deadline mechanisms and five-plus overlapping "can we save right now?" trackers on top of the queue instead of trusting it. **Six of the seven inline findings live in the seams between these parallel mechanisms.** Collapsing them makes all seven findings disappear structurally (no spot-patches needed) and cuts roughly 400–500 of the ~1,100 new production lines.

### Keep as the backbone

- `runDocumentWrite` / `documentWriteTail` — the 12-line promise-chain queue is the right primitive; everything below leans on it harder.
- Routing saves/restores/resets/compaction/bridge-applies/admin mutations through the queue, including re-reading `resetEpoch`/relationships after acquiring it.
- `bridgeAuth.ts` — small, single-purpose, matches the stated goal.
- Admin cleanup mutating live room state instead of `loadDatabasePlayData` — deletes a real lost-update bug.
- The generation check guarding `commitCompactedDocument`'s live-doc replacement — WebSocket Y-updates bypass the queue, so this one defends a real interleaving.

### Collapse

**1. Write-availability state → one flag; delete the `isSkippingSave` setter** (eliminates finding 6, ~50–70 lines)
The 5-kind `DocumentWriteState` enum is barely consumed: every consumer reduces to `kind !== "idle"` once `recovery-required` is gone (see #2), and bridge applies queue behind the tail anyway. Production code never calls the `isSkippingSave` setter — only tests do — and it force-idles the state machine. A single `documentMaintenanceInProgress` boolean suffices; update the 4 test call sites.

**2. Delete the `pendingDocumentReset` durable marker + reconcile + retry machinery** (~170–200 lines)
The crash window it protects (snapshot committed, epoch not yet in DO storage) is already covered twice: every reset snapshot embeds its epoch in the document itself before the Supabase write, and both the load path and `getAutosaveResetEpochDecision`'s `promote-server-epoch` path promote it. Also, `setResetEpochAfterDocumentCommit`'s 3-attempt retry "backoff" is `await Promise.resolve()` — a microtask yield that fails identically all 3 times. Let the error propagate; the doc-embedded epoch guarantees convergence.

**3. Unify transient-persistence recovery with the existing load-backoff ledger** (eliminates findings 1 and 7, ~170–200 lines)
`persistenceRecoveryPending` + `persistenceRecoveryAttempts` + `persistenceRecoveryRetryAfter` is a second parallel backoff ledger inside the circuit breaker — and `recoveryPending === true` is nearly equivalent to `recoveryRetryAfter !== null`, so even the new state is redundant with itself. Because recovery "owns its own deadline," it must bypass load backoff and suppress quarantine, which is exactly what creates the unbounded OOM crash loop (finding 1). Instead: reuse the existing `loadRetryAfter`/failure-count backoff, plus one durable bit meaning "next successful hydration must restore-with-bumped-epoch and kick clients." Recovery loads then flow through `beginRiskyOperation("load")` like any other load — OOMs accrue evidence, backoff applies, quarantine fires. `enterTransientPersistenceMode` goes back to being synchronous in-memory work, fixing finding 7.

**4. Shrink `documentSaveRetry` to a single durable `retryAt`, consume-or-reschedule** (eliminates findings 2 and 3, ~35–50 lines)
The "doc changed during upsert → schedule retry" branch duplicates y-partyserver's debounced `onSave`, which already fires again after any doc update — so counting a mid-save change as a failure attempt (the max-backoff ratchet, finding 3) buys nothing. Clear the retry unconditionally on a successful upsert. With one path — alarm fires → delete the key → attempt save → on throw, write a fresh `retryAt` — the "handled = consumed" invariant is trivially true and the silent-skip infinite alarm loop (finding 2) becomes unrepresentable.

**5. Encode inside the queue; delete the re-encode staleness guard** (eliminates finding 4, ~15 lines)
The guard exists to protect the one caller (admin `force-save-live`) that encodes before entering the queue — an ordering the endpoint itself chooses. Move the encode inside the queued closure (or have force-save-live call `persistLiveDocument`) and "stale" is unrepresentable; the public wrapper no longer needs a base64 parameter at all.

**6. Inline `bridgeRequestPolicy.ts`; restore delta-based `changed` in the lease merge** (eliminates finding 5, ~90–130 lines incl. tests)
`getBridgeApplyRelationship` wraps two `.includes()` checks that were 6 inline lines on main; the 403-on-unknown behavior is a 4-line early return at the call site. The lease-policy extraction rewrote main's correct delta-based `changed` computation into "changed for every non-empty request" (the O(K²) re-hydration, finding 5). The intended lease-renewal feature needs only ~6 lines on the original inline code: report changed when membership actually changed or when `lastSeen` is older than a renewal threshold.

**7. (Minor) Flatten `mutateAdminPlayData`'s double-layer result unions** (~20–30 lines) — `commit|skip` mapped into `committed|skipped` is two discriminated unions for what both call sites consume as "did it commit."

### Net effect

| # | Collapse | Findings eliminated | Lines saved |
|---|---|---|---|
| 1 | Write-state enum + setter → one flag | 6 | ~50–70 |
| 2 | Pending-reset marker → doc-embedded epoch promotion | (defuses 6's blast radius) | ~170–200 |
| 3 | Parallel recovery ledger → existing load backoff + one bit | 1, 7 | ~170–200 |
| 4 | Save retry → single retryAt | 2, 3 | ~35–50 |
| 5 | Staleness guard → encode in queue | 4 | ~15 |
| 6 | Inline bridge policy files, fix lease delta | 5 | ~90–130 |
| 7 | Flatten admin result types | — | ~20–30 |

`party.ts`'s +581-line growth shrinks to roughly +150–200 lines centered on the write queue, bridge auth, and a slimmed recovery bit — and every inline finding is fixed by deletion rather than by more machinery.

---

partykit/party.ts
```diff
@@ -2819,9 +3039,30 @@ export class PartyServer extends YServer {
 
   // PartyKit Alarm: invoked when storage alarm rings
   override async onAlarm(): Promise<void> {
+    const loadDeferred = await this.circuitBreaker.getLoadDeferredResponse();
+    if (loadDeferred) {
```
spencerc99, line 3053: **Alarm spin loop: a load-deferred room with a past-due `documentSaveRetry` hot-loops here.** This early return re-arms via `scheduleNextAlarm()`, which takes `min(..., documentSaveRetry.retryAt, ...)` and will happily arm a past timestamp (the previous alarm was consumed when this handler ran). But the branch that consumes `documentSaveRetry` is below the early return — unreachable while the load is deferred.

Scenario: Supabase outage → warm room's autosave fails → `documentSaveRetry{retryAt: now+1s}` written durably → isolate evicted → the retry alarm wakes it cold → load fails → `deferFailedLoad` defers for 60s (growing to hours on the backoff ladder) → alarm fires at the past `retryAt` → early return → re-arm at the same past `retryAt` → immediate re-fire, tight-looping for the entire deferral window. This reintroduces the alarm-spin class the rework was meant to eliminate.

Fix: consume (or push forward) a due `documentSaveRetry` in this deferred branch. It's meaningless after an isolate restart anyway — the unpersisted doc died with the isolate. The new "a deferred room does not read the document when its alarm fires" test doesn't cover this (no saveRetry present); add one with a stale due retry.

---

partykit/party.ts
```diff
@@ -1030,44 +1144,48 @@ export class PartyServer extends YServer {
...
           `[Restore Snapshot] Database save failed:`,
           getErrorMessage(saveError),
           saveError
         );
-        throw new Error(`Failed to save snapshot: ${getErrorMessage(saveError)}`);
+        throw new Error(
+          `Failed to save snapshot: ${getErrorMessage(saveError)}`
+        );
       }
       console.log(`[Restore Snapshot] Successfully saved snapshot to database`);
```
spencerc99, line 1190: **Warm-path regression: if `setResetEpoch` throws after this upsert succeeds (isolate alive), the next autosave silently reverts the committed restore.** The restore rejects, `finally` clears `documentMaintenanceInProgress`, and the live doc + DO epoch remain old-but-mutually-consistent — so `getAutosaveResetEpochDecision` can't detect the boundary, and the next user write's autosave overwrites the restored snapshot with the old document.

The deleted machinery (`setResetEpochAfterDocumentCommit` retry + pending marker + save block) covered exactly this warm-failure window; the new fault-injection tests only cover the isolate-*death* variant, where the next load converges via the doc-embedded epoch. The analogous compaction case is benign (reverting to the uncompacted doc loses nothing), and the operator does see the 500 and can retry — so this is a deliberate-tradeoff candidate rather than a must-fix.

Two acceptable resolutions: (a) accept and document — the operator retries on error, isolate death self-heals; or (b) on epoch-write failure after a committed restore, fail closed instead of open: keep saves blocked (e.g. leave `documentMaintenanceInProgress` set or abort the DO) so the room reloads and converges via the embedded epoch, rather than resuming autosaves against the stale live doc. (b) is a few lines and restores the old guarantee without resurrecting the marker machinery.

---

spencerc99 (commented):
Re-review of the rework commit 0218fc4 ("partykit: simplify persistence recovery") against the agreed plan. Verified by reading the full diff and running the suite in a clean worktree: 281 pass / 0 fail (was 274), `check:partykit` clean, no new stderr noise.

**All 7 original findings are closed, each by the intended deletion:**
1. Recovery OOM loop — recovery attempts/retryAfter ledger deleted; recovery loads flow through `beginRiskyOperation("load")` + `deferFailedLoad`; both quarantine-suppression guards removed (new test: "pending recovery cannot suppress quarantine at the load threshold").
2. Silent-skip alarm loop — retry is consume-first (`clearDocumentSaveRetry()` before the attempt), rescheduled only on throw.
3. Backoff ratchet — retry cleared unconditionally on successful upsert; generation compare removed from autosave (kept in `commitCompactedDocument`, correctly).
4. force-save-live staleness 500 — `saveDocumentBase64` + re-encode guard deleted; encode happens inside the queued closure; new test covers a write landing during the queue wait.
5. O(K²) re-hydration — delta-based `changed` restored with a 6h renewal threshold; no-op re-registrations skip `subscribeAndHydrate`.
6. `isSkippingSave` setter — deleted; enum collapsed to `documentMaintenanceInProgress` with the save-permission matrix preserved.
7. Transient-mode entry — synchronous in-memory again; durable writes in try/catch (new test: "transient degradation survives a recovery-marker storage failure").

The `pendingDocumentReset` marker was fully deleted with the gating fault-injection tests present and passing (restart convergence via doc-embedded epoch, both restore and compaction paths). The P2 stderr capture was done for the named test; two pre-existing uncaptured traces elsewhere in `quarantineLoad.test.ts` remain (same as before this commit).

**Two new findings from this pass, inline:** an alarm hot-loop when a load-deferred room holds a past-due save retry (party.ts:3043 — small targeted fix), and a warm-path window where a failed epoch write after a committed restore lets the next autosave revert it (party.ts:1190 — needs an accept-and-document vs. fail-closed decision).

**Deliberate behavior changes worth a maintainer sign-off** (consequences of the agreed unification, not bugs): (a) Supabase outages now accrue quarantine evidence — an outage of roughly 7+ hours quarantines rooms, which then need a manual quarantine-clear after recovery (previously outages could never quarantine); (b) save retry is now a fixed 1s delay with no cap, so during a save-only outage every ready room upserts once per second indefinitely (was capped exponential at 60s) — consider a gentler constant or small cap.

---

partykit/roomCircuitBreaker.ts
```diff
@@ -337,6 +332,30 @@ export class RoomCircuitBreaker {
...
+      failureCount,
+    });
+    this.loadDeferredUntil = result.quarantined ? null : result.retryAt;
+  }
+
+  /**
+   * A returned load error proves the isolate survived, so it is not evidence
+   * of the vanished work this circuit breaker quarantines.
+   */
+  async deferObservedLoadFailure(): Promise<void> {
```
spencerc99, line 347: **The observed-outage deferral doesn't survive isolate eviction — cold wakes retry immediately, ignoring the stored deadline.** `completeRiskyOperation("load")` deletes `quarantineLoadAttempts`, and `shouldDeferLoad()` early-returns `false` when that count is 0 — before ever reading `loadRetryAfter`. The warm-path gate (`getLoadDeferredResponse`) only checks the in-memory `loadDeferredUntil`. So the `loadRetryAfter` this method writes is never read on a cold start:

Outage → observed failure → this method runs (count zeroed, retryAt stored, memory deferral set) → isolate evicted (likely mid-outage; the alarm wakes it cold) → `shouldDeferLoad()` sees 0 failures → immediate full reload attempt (3 timeouts, ~15s) → observed failure again → repeat on every wake, with the intended backoff never applying across restarts. Clients reconnecting during the outage each trigger the same cycle.

Suggested fix, keeping the single ledger: make `shouldDeferLoad()` honor a stored, not-yet-due `loadRetryAfter` even when the failure count is 0 (defer + set `loadDeferredUntil`), and delete the key when it is due so the load proceeds. That keeps observed outages non-quarantining (count stays 0) while making the deferral durable. Worth a test: observed failure → fresh circuit breaker instance (simulated cold start) → `shouldDeferLoad()` returns true until `retryAt` passes.

Minor, same method: `failureCount: 1` is fixed, so repeated observed outages never escalate past base backoff — probably fine (flat, bounded retries during outages), but worth a comment saying it's intentional.

---

spencerc99 (commented):
Reviewed a2915d1 ("separate outages from load crashes"). Direction is right and it settles both flagged sign-offs: observed Supabase failures no longer accrue quarantine evidence (only vanishes/OOMs do), and the save retry moves from 1s to 60s. Suite green in a clean worktree (283 pass, tsc clean). This also front-runs part of the #424 port — good.

One finding, inline: the observed-outage deferral is in-memory only in practice — `completeRiskyOperation` zeroes the failure count, `shouldDeferLoad` early-returns on count 0 without reading `loadRetryAfter`, so a cold-started isolate retries immediately on every wake for the whole outage. Fix is small (honor a stored not-yet-due `loadRetryAfter` at count 0) and deserves a cold-start regression test.

---

partykit/roomCircuitBreaker.ts
spencerc99: Reviewed 0bb4140 (suite green, 284 pass): it addresses the client-facing half of this — `getClientLoadDeferredResponse` correctly keeps outage rooms serving transient connections during the backoff window (and it reads the stored `loadRetryAfter` at count 0, so the client path is durable). Good fix for the regression a2915d1 introduced there.

The server-side half remains as described: `shouldDeferLoad()` still ignores the stored deadline at count 0, so a cold-started isolate (eviction mid-outage) runs one eager ~15s load-attempt cycle per isolate start instead of waiting out the backoff. Bounded — once per isolate lifetime, and the alarm wake at `retryAt` is legitimate — but it means the first visitor after each eviction during an outage stalls ~15s on connect (onLoad runs before the connection is served), where the durable guard would give them instant transient service at the cost of up to one backoff interval of delayed recovery detection.

If the eager cold-start attempt is the intended tradeoff (fast recovery when a human shows up), a one-line comment in `deferObservedLoadFailure` saying so is enough and I'll consider this closed. Otherwise the guard is: in `shouldDeferLoad`, when count is 0 but a stored `loadRetryAfter` is not yet due, defer (and set `loadDeferredUntil`); delete the key when due. Non-blocking either way.

---

spencerc99: be869f9 closes this, and more cleanly than my suggested guard: `shouldDeferLoad` at count 0 now sets `loadDeferredUntil` from the stored deadline but returns false — so the room still starts up and serves transient connections instantly, while `onLoad`'s `isLoadDeferred()` check skips the eager Supabase read. Cold wake mid-outage: zero document reads, no 15s connect stall, no quarantine evidence, and the alarm at `retryAt` still performs the scheduled retry. The new restart test pins exactly that. Verified: 285 pass, tsc clean. Resolving.

---

PR #421, opened 2026-08-26
Extension: index scrap deduplication

spencerc99:
## Summary

- persist each valid scrap's canonical identity in IndexedDB and index it with a non-unique `canonicalScrapKey` index
- remove the MV3 worker's lazy full-archive scan and keep canonical existence checks plus accepted writes in one read-write transaction
- pre-deduplicate incoming batches and update aggregates only for accepted event rows

## Migration

IndexedDB schema v12 creates the non-unique index and backfills canonical keys for existing `element` rows once during upgrade. The migration does not delete or collapse historical duplicates; every duplicate remains stored and indexed. A non-unique index lets any retained duplicate continue to block later copies. IndexedDB automatically removes index entries when clear or prune deletes the underlying event.

The canonical field remains storage-internal and is stripped from queries/uploads. If canonical identity semantics change later, the stored keys will require another schema migration.

## Forced MV3 restart benchmark

Real Chrome MV3 artifact, extension-origin IndexedDB, forced service-worker target close, then first `STORE_EVENTS` response; three isolated-profile runs per scale.

| Stored scraps | Before | After |
| --- | --- | --- |
| 0 | 19.886–23.639 ms | 26.503, 29.190, 31.267 ms |
| 1,000 | 20.593–21.277 ms | 12.083, 12.088, 13.117 ms |
| 10,000 | 108.644–113.462 ms | 27.109, 29.689, 30.206 ms |

The first-write cost no longer grows with the 10,000-row archive. All nine writes returned `{ success: true }`.

## Tests

- `bun run build-packages`
- `bun run test:extension` — 136 files, 883 tests
- `bun run check:extension`
- `bun run build-extension`
- focused real IndexedDB coverage for fresh stores, v11 archive migration, preserved duplicate rows, same-batch duplicates, concurrent transactions, canonical identity semantics, prune/clear lifecycle, internal-field stripping, and aggregate counts

No screenshot is included because this changes no visual surface. No `PENDING.md` entry is included because this is an internal storage performance fix rather than a user-facing release-note item.


spencerc99:
## Code review

- **P1 — Preserve the canonical backfill when skipping older schema versions** ([LocalEventStore.ts:602-653](https://github.com/spencerc99/playhtml/blob/ed0a6fafd2cf6215e1ea7950eeae50c7ff7b89a5/extension/src/storage/LocalEventStore.ts#L602-L653))

  The new version-12 cursor and the existing version-9 upload-state cursor can both read and update the same event during one upgrade. For databases at version 8 or earlier, the later upload-state update writes the stale row it read before the canonical update, removing canonicalScrapKey. I reproduced this with a valid version-8 database containing one image scrap: after LocalEventStore upgraded it to version 12, the row had uploadState=pending but the canonicalScrapKey index count was 0. That archived scrap therefore does not block the next duplicate, contrary to the migration guarantee. Combine these field migrations into one cursor update, or explicitly sequence the canonical backfill after the earlier cursor completes, and cover a skipped version-8-to-12 upgrade.

PR #424, opened 2026-08-26
partykit: recover Supabase hydration outages

spencerc99:
## Summary

- build on #416's single authoritative recovery path and existing `loadRetryAfter` state
- close new, existing, and hibernated WebSocket clients with 1013 `Room Loading` while hydration recovery is pending
- treat returned Supabase timeout/HTTP failures as provider outages without adding quarantine evidence, while preserving any earlier vanished-attempt evidence
- retry provider outages after about 5 seconds with a 5-minute cap, and keep #416's one-minute save retry instead of the earlier one-second interval
- update the transient outage smoke to require 1013 client gating

## Rationale

#416 now owns durable admission, alarm scheduling, authoritative reload, reset boundaries, and the single risky-operation ledger. This follow-up only changes client admission and failure bookkeeping within that mechanism. It does not add recovery counters or storage keys.

A caught Supabase error proves the isolate survived, so the current load attempt is released from quarantine accounting and a short recovery deadline is scheduled. A vanished attempt still accrues evidence exactly as #416 defines. Existing crash evidence is preserved when a later provider outage is caught.

The product behavior is intentionally fail-closed during startup hydration: clients reconnect after receiving 1013 instead of using awareness against an empty placeholder document.

## Verification

- failing-first regression run for new-connection gating, hibernated-socket gating, three failed reads followed by alarm recovery, and outage/crash ledger separation
- `bun run check:partykit`
- `bun test partykit/__tests__/` (286 passed)
- `node --check smoke-tests/partykit/supabase-transient-smoke.mjs`
- local Wrangler outage smoke with an unreachable Supabase endpoint: both clients received 1013 `Room Loading`, and the admin write returned 503 `room_load_deferred`
- `git diff --check`

## Release notes

No changeset or public documentation update. This changes internal PartyKit recovery behavior and its operational smoke test, not the published package API or user-facing setup.

### DIFF
--- a/partykit/roomCircuitBreaker.ts
+++ b/partykit/roomCircuitBreaker.ts
@@ -40,6 +40,8 @@ type RoomCircuitBreakerOptions = {
     failureThreshold: number;
     failureBackoffMs: number;
     failureBackoffMaxMs: number;
+    observedLoadFailureBackoffMs: number;
+    observedLoadFailureBackoffMaxMs: number;
   };
   activateTransientPersistence: (quarantine: QuarantineState) => void;
   startRealtimeSync: () => Promise<void>;
@@ -116,6 +118,20 @@ export class RoomCircuitBreaker {
     return this.options.readPositiveNumber(
       "FAILURE_BACKOFF_MAX_MS",
       this.options.defaults.failureBackoffMaxMs
+    );
+  }
+
+  private getObservedLoadFailureBackoffMs(): number {
+    return this.options.readPositiveNumber(
+      "SUPABASE_RECOVERY_RETRY_DELAY_MS",
+      this.options.defaults.observedLoadFailureBackoffMs
+    );
+  }
+
+  private getObservedLoadFailureBackoffMaxMs(): number {
+    return this.options.readPositiveNumber(
+      "SUPABASE_RECOVERY_RETRY_MAX_DELAY_MS",
+      this.options.defaults.observedLoadFailureBackoffMaxMs
     );
   }
 
@@ -290,6 +306,18 @@ export class RoomCircuitBreaker {
     await this.storage.delete(this.retryKeyFor(kind));
   }
 
+  private async releaseLoadAttempt(loadAttempts: number): Promise<void> {
+    const remainingAttempts = loadAttempts - 1;
+    if (remainingAttempts <= 0) {
+      await this.storage.delete(STORAGE_KEYS.quarantineLoadAttempts);
+      return;
+    }
+    await this.storage.put(
+      STORAGE_KEYS.quarantineLoadAttempts,
+      remainingAttempts
+    );
+  }
+
   private async handleRepeatedFailures({
     kind,
     failureCount,
@@ -344,12 +372,12 @@ export class RoomCircuitBreaker {
    * A returned load error proves the isolate survived, so it is not evidence
    * of the vanished work this circuit breaker quarantines.
    */
-  async deferObservedLoadFailure(): Promise<void> {
-    await this.completeRiskyOperation("load");
+  async deferObservedLoadFailure(loadAttempts: number): Promise<void> {
+    await this.releaseLoadAttempt(loadAttempts);
     const retryAt = getFailureRetryAt({
-      failureCount: 1,
-      baseMs: this.getFailureBackoffBaseMs(),
-      maxMs: this.getFailureBackoffMaxMs(),
+      failureCount: loadAttempts,
+      baseMs: this.getObservedLoadFailureBackoffMs(),
+      maxMs: this.getObservedLoadFailureBackoffMaxMs(),
       now: Date.now(),
     });
     await this.storage.put(STORAGE_KEYS.loadRetryAfter, retryAt);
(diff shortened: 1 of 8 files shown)


spencerc99:
## Code review

partykit/roomCircuitBreaker.ts
```diff
@@ -344,12 +372,12 @@ export class RoomCircuitBreaker {
...
    */
-  async deferObservedLoadFailure(): Promise<void> {
-    await this.completeRiskyOperation("load");
+  async deferObservedLoadFailure(loadAttempts: number): Promise<void> {
+    await this.releaseLoadAttempt(loadAttempts);
     const retryAt = getFailureRetryAt({
-      failureCount: 1,
-      baseMs: this.getFailureBackoffBaseMs(),
-      maxMs: this.getFailureBackoffMaxMs(),
+      failureCount: loadAttempts,
```
spencerc99: **The exponential backoff never engages for a pure outage — retries sit flat at the 5s base for the whole outage.** `releaseLoadAttempt` runs first, so on the next attempt `beginRiskyOperation` returns 1 again; `failureCount: loadAttempts` is therefore 1 on every consecutive observed failure (it only exceeds 1 when *vanish* evidence happens to be accumulating). The 5s→5min ladder these constants promise reduces to a constant 5s.

Concretely, during a full Supabase outage each room cycles: load attempt (3 × 5s timeouts ≈ 15s) → retryAt = now + 5s → alarm/reload → attempt again — i.e. near-continuous load attempts, 3 HTTP requests every ~20s per room, for the entire outage. Multiplied across live rooms that's sustained pressure on a database that's already struggling, and it's also ~12x more aggressive than what merged in #416 (flat 60s base). The PR summary says "bounded exponential backoff," which is the right goal — this just doesn't implement it yet.

One single-ledger way to get real escalation without a failure counter: derive the streak from the deadline you already store. On observed failure, read the existing `loadRetryAfter`; if it's in the past (a due retry that failed again), double the previous delay — e.g. store `retryAt` and compute `previousDelay = retryAt_prev - (time it was set)` … since only `retryAt` is stored, the simplest robust form is: `newDelay = clamp(base, 2 * (Date.now() - retryAt_prev), max)` when `retryAt_prev` exists and is past (time since the last scheduled retry approximates the last delay), else `base`. Or, frankly, a flat base of 30–60s with the 5min cap dropped would also be fine and simpler — the important property is not hammering a down database every 20s. Worth a test either way: two consecutive observed failures → second `retryAt` delay strictly greater than the first (or at the agreed flat value).

---

spencerc99 (commented):
Reviewed the rework (9297235, force-pushed onto the #416 merge). This matches the agreed port plan and is the right shape: no duplicate recovery ledger (zero new storage keys), the 1013 client gating reads the circuit breaker's single deferral state and is applied consistently (onConnect, onMessage, HTTP 503, and closing existing connections on a failed hydration), transient-mode messaging updated to the new connection policy, and the smoke test rewritten for it. Verified locally: 286 pass / 0 fail, `check:partykit` clean, and no merge conflicts against current main (b0efc81+). Two things are genuinely better than the interim state: `releaseLoadAttempt` decrements only this start's increment, so prior vanish/OOM evidence is preserved instead of wiped on an observed outage; and outage backoff got its own env-tunable constants.

One finding, inline: the promised 5s→5min exponential never escalates in practice — the attempt is released before the delay is computed, so `failureCount` is 1 on every consecutive observed failure and rooms retry a down Supabase every ~20s (15s of timeouts + 5s wait) for the whole outage. Suggested fixes in the comment (derive the streak from the stored deadline, or just use a flat 30–60s base); either resolves it.

Also note the locked-in product decision, deliberate per the plan but worth stating once: during an outage clients are now closed with retryable 1013 instead of receiving awareness-only transient service — the behavior #416's staging matrix asserted has intentionally changed.

---

partykit/roomCircuitBreaker.ts
spencerc99: 60e7aa1 fixes a real but different problem: an observed failure no longer *shortens* a stricter existing deadline (the guarded-retain case, now tested — good). The original finding still stands, though: when there is no stored deadline, or the stored deadline is in the past — which is exactly the consecutive-outage-failure case, since the due retry just ran and failed — `max(firstRetryAt, guardedRetryAt)` resolves to `now + 5s`. So a pure outage still cycles at ~20s (15s of timeouts + 5s wait) per room for its whole duration; suite confirms (287 pass, but no test pins the consecutive-failure cadence).

Within the structure you now have, the simplest complete fix is to raise the flat base: set `DEFAULT_SUPABASE_RECOVERY_RETRY_DELAY_MS` to 30–60s (keeping the env override and the 5min clamp). Escalation-from-the-stored-deadline is also possible (`guardedRetryAt` in the past → delay = clamp(base, 2 × (now − guardedRetryAt), max)), but it converges to ~2× the attempt duration rather than a clean ladder — the flat 30–60s base is simpler and achieves the property that matters: not hammering a down database every 20 seconds. Plus one test: two consecutive observed failures → the second `retryAt` is ≥ the agreed cadence.

---

spencerc99: c610789 resolves this. Pre-arming the next deadline (crash-ladder base, ~60s) before a due retry's attempt means a consecutive observed failure retains it through the guarded max — so the cadence is one fast 5s first retry, then flat ~60s per attempt for the rest of the outage, well within the 30–60s bar. The pre-arm being durable before the risky attempt also protects the count-0 path against a vanish mid-attempt, and success still clears the key via completeRiskyOperation. The new restarted-outage test pins the cadence and the absence of quarantine evidence. Verified: 288 pass, tsc clean. Resolving.

---

PR #422, opened 2026-08-26
extension: lazy-load rare content-script UI

spencerc99:
## Summary

- keep cursor, navigation, viewport, keyboard, and gated scrap collectors on their existing all-page initialization paths
- move the historical overlay, milestone toast, toast styles, and React Shadow DOM mounting support into an on-demand WXT entrypoint
- load that entrypoint only when a history or milestone event actually displays the rare UI, preserving replacement, teardown, error recovery, and stale-request handling
- add a headless paired benchmark that stops only after all four normal collectors report enabled

## Content-script size

Measured from same-machine production Chrome MV3 builds of current main (`769bba1c`) and this branch (`59951aa1`):

| Encoding | Main | This PR | Change |
| --- | ---: | ---: | ---: |
| Raw | 1,203,388 B | 1,120,542 B | -82,846 B (-6.88%) |
| gzip -9 | 516,378 B | 489,429 B | -26,949 B (-5.22%) |
| Brotli q11 | 303,110 B | 280,653 B | -22,457 B (-7.41%) |

The on-demand `content-page-ui.js` artifact is 340,207 B raw, 107,710 B gzip, and 92,587 B Brotli. This trades total package bytes for less code parsed and initialized on every supported page. Site-specific React features still leave some React code in the eager content bundle; broader feature splitting is outside this focused change.

## Collector-ready startup

I ran 15 alternating same-machine pairs in headless Chrome. Every run used a fresh profile, started and seeded the extension worker before timing, navigated to the same local no-script page, and stopped when cursor, navigation, viewport, and keyboard collectors all reported enabled. The rare UI script was not requested during startup in any run.

| Metric | Main median | This PR median | Paired median change | Lower-work PR pairs |
| --- | ---: | ---: | ---: | ---: |
| Collector-ready wall time | 96.292 ms | 94.430 ms | -1.813 ms | 11/15 |
| TaskDuration | 66.433 ms | 64.506 ms | -1.916 ms | 14/15 |
| ScriptDuration | 32.058 ms | 31.683 ms | -0.389 ms | 13/15 |
| JSHeapUsedSize delta | 3,124,644 B | 3,016,264 B | -108,380 B | 15/15 |

[...]

PR #427, opened 2026-08-27
extension: remove startup PlayHTML capability scan

spencerc99:
## Summary

- stop scanning every page for PlayHTML capability attributes during content-script initialization
- remove the now-unreachable startup scan method
- preserve the on-demand CHECK_PLAYHTML_STATUS path and its live DOM count
- add executable content-entrypoint coverage that proves startup skips the selector while an explicit status request sees elements added after initialization

## Rationale

The startup result was discarded, so every supported page paid for a full capability-selector query without changing extension behavior. The popup status request still needs a current count and continues to scan on demand.

## Verification

- bun run build-packages
- focused regression test, including a pre-fix failure on the startup selector
- bun run test:extension: 141 files and 897 tests passed
- bun run check:extension
- production Chrome MV3 build
- bun run smoke:extension: service worker, content script, and popup loaded in isolated headless Chrome
- git diff --check
- final conservative diff review: no findings

## Scope and overlap

PR #422 was checked and is closed without this cleanup. Current open extension PRs were checked; #264 also touches content.ts for an unrelated auth challenge bridge and does not duplicate this removal.

No screenshots because visible UI behavior is unchanged. No extension/PENDING.md entry because this is internal startup maintenance. No package docs, starter templates, or changeset are needed because published package behavior and APIs are unchanged.

### DIFF
--- a/extension/src/__tests__/content-playhtml-status.test.ts
+++ b/extension/src/__tests__/content-playhtml-status.test.ts
@@ -1,0 +1,124 @@
+// ABOUTME: Verifies PlayHTML capability scans run only for explicit status requests.
+// ABOUTME: Exercises content-script initialization and status messaging against the real DOM.
+
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const storageGet = vi.hoisted(() => vi.fn());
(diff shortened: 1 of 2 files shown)


spencerc99:
## Code review

Issue #289, opened 2026-07-09
API redesign proposal: unified data model (buckets, addresses/handles, users/live split)


spencerc99:
Proposal from a full audit of the public API surface (2026-07-07). This consolidates the data model, authoring surface, and cross-page/cross-site story into a small set of composable primitives, with an incremental, migration-free rollout path.

## Motivation

The current surface has grown by accretion. Mapping every primitive onto two axes — **ownership** (shared: one value for the room, vs per-user: each person publishes their own) and **lifetime** (persisted / ephemeral / local) — shows the problem at a glance:

| | Shared (one value) | Per-user (map of values) |
|---|---|---|
| **Persisted** | element `setData` + `createPageData` — 2 names, 1 primitive | gap — hand-rolled via pid-keyed writes into shared data |
| **Ephemeral** | gap — in practice always derived or an event | element awareness + page presence + cursors — 3 APIs, 3 transports |
| **Local** | `localData` (in-memory only) | identity (localStorage + multiple broadcast write sites) |

Six-plus names for four cells; one cell covered three times. Scope (element vs page vs room) is not a third kind of thing — it is just a key: `createPageData` already stores under a synthetic `__page__` tag through the same element-proxy machinery.

Additional findings from the audit: `globalData` is dead code; `elementHandlers`/`eventHandlers`/`sharedPermissions` are raw mutable maps exposed publicly while `syncedStore` gets a read-only proxy; three teardown verbs (`removePlayElement` / `deleteElementData` / `unregister`) for two concepts; `extraCapabilities` and `define()` register into the same map through different validation paths; deprecated aliases (`additionalSetup`, `hasSynced`) still live.

## The model

Four things, one of which is free:

1. **`data`** — shared state. One value per key, everyone reads/writes, persisted.
2. **`live`** — per-user state (today's element awareness, renamed). Each person writes their own slot; everyone reads all slots; cleaned up mechanically on disconnect. Per-user ownership of ephemeral state is not a Yjs artifact: disconnect cleanup requires an owner, which is why every multiplayer system (Liveblocks, Phoenix Presence, Figma) converges here. Note: `live` is the ephemeral bucket **on each user** (at page scope, reached through `playhtml.users`; at element scope, through the element ctx) — it is not a competing top-level roster API.
3. **`users`** — the roster and ownership model: durable identity (name / color / custom profile traits). Being built in #284. users = who, presence/live = what they're doing, cursors = a rendering layer on top.
[...]

PR #430, opened 2026-08-27
Add element live users and update APIs

spencerc99:
## Summary

- add `live`, `users`, `setLive`, `getLive`, and `getUsers` for element-scoped per-user data
- add `update` as the preferred imperative renderer and run it for shared data and live-user changes
- keep the awareness and `updateElement` spellings as deprecated compatibility aliases, with errors when both alias names are supplied
- expose the same live-user model through React render props
- update public docs, both starters, the element-building skill, tests, and changesets

## Why

Element-scoped presence previously required separate awareness callbacks and raw value collections. The new API gives each renderer one model: `live` is my ephemeral value, `users` joins every current value with its user identity, and `update` renders both shared data and user changes.

This is the core element naming/render slice from #289. Page-level Users live channels, cursor renderer changes, data/useData, read tracking, drag/onClick cleanup, awareness fallback removal, and teardown renames remain out of scope.

## Verification

- `bun run -C packages/playhtml test` equivalent from the package directory: 538 tests passed
- `bun run -C packages/react test` equivalent from the package directory: 57 tests passed, including type and real-core integration tests
- `bun run build-packages`
- `bun run lint`
- `bun run -C apps/docs build` equivalent from the docs directory
- `bunx vite build` in `templates/react-starter`
- real-browser verification of the docs, React starter, and a temporary HTML starter using the locally built unreleased package
- both starter hover examples changed from `0 here` to `1 here`; fresh starter tabs had no console errors

Part of #289.

### DIFF
--- a/.changeset/bright-users-update.md
+++ b/.changeset/bright-users-update.md
@@ -1,0 +1,7 @@
+---
+"@playhtml/common": minor
+"playhtml": minor
+"@playhtml/react": minor
+---
+
+Add the element `live`, `users`, `setLive`, and `update` APIs so one renderer can combine shared state with current per-user values and identity. Existing awareness and `updateElement` names remain available as deprecated compatibility aliases.
--- a/apps/docs/src/content/docs/advanced/dynamic-elements.mdx
+++ b/apps/docs/src/content/docs/advanced/dynamic-elements.mdx
@@ -46,7 +46,7 @@ document.body.appendChild(note);
 
 playhtml.register(note, {
   defaultData: { text: "" },
-  updateElement: ({ element, data }) => {
+  update: ({ element, data }) => {
     element.textContent = data.text;
   },
 });
--- a/apps/docs/src/content/docs/advanced/merging-data.md
+++ b/apps/docs/src/content/docs/advanced/merging-data.md
@@ -139,7 +139,7 @@ setData((draft) => {
 
 ## Reactive callbacks
 
-The riskiest conflict is a callback that reads shared data, writes shared data, and re-runs when that same data changes. React effects, subscriptions, and vanilla `updateElement` callbacks can all do this.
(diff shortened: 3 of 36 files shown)


spencerc99:
## Preview evidence

**Element live/users documentation and both starters use one per-user rendering model.**

[Open preview](http://127.0.0.1:4321/docs/custom-elements/) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-mtb6l3ao-f44cd85f) · Commit `25829f5`

### Vanilla live users guide

Vanilla live users guide

The custom-elements guide now explains live, users, setLive, and one update renderer with a complete vanilla HTML example.

[Open captured page](http://127.0.0.1:4321/docs/custom-elements/) · 1280×720

### React starter live hover

React starter live hover

Hovering the React reaction button publishes live state and renders one current user.

[Open captured page](http://127.0.0.1:4322/) · 1280×720

### Vanilla starter live hover

Vanilla starter live hover

The vanilla starter uses the same live/users API and renders one current user while hovered.

[Open captured page](http://127.0.0.1:4323/) · 1280×720

spencerc99:
## Code review

- **High: Element `users` can stay empty or stale when identity arrives after live data** ([`packages/playhtml/src/index.ts:1971`](https://github.com/spencerc99/playhtml/blob/25829f5f6f3118db92cf7b4f32b2d88783937e6a/packages/playhtml/src/index.ts#L1971)): `getElementUsers` joins the live-value map against `usersAPI.getAll()` only while an element-awareness update is rendering. The users module receives Yjs identity changes independently, and this PR does not subscribe element renderers to `usersAPI.onChange`. The presence socket can therefore deliver a remote element value before the Yjs identity state. That render omits the user, and the later identity arrival does not render the element again, so the user stays missing until some unrelated live value changes. The same ordering can leave a changed name or color stale. Refresh the element render path when the users snapshot changes, or join identity and live data from one authoritative feed. Add coverage where socket live data arrives before the Yjs identity.

spencerc99:
## Code review

PR #438, opened 2026-08-28
partykit: preserve slow document hydration

spencerc99:
## Summary

- give Supabase document hydration one shared 15-second deadline instead of aborting each read at 5 seconds
- record slow and retried hydration timing without logging document data
- add real loopback regression coverage for a response that arrives after 5 seconds

## Rationale

Production telemetry showed document reads repeatedly hitting the previous 5-second cutoff even when Supabase could still return the room. A shared deadline preserves explicit-error retries while allowing a slow viable read to complete.

This keeps `@supabase/supabase-js`; it does not change the database client or dependencies.

## Testing

- `bun test partykit/__tests__` (289 passed)
- `bun run check:partykit`
- Wrangler production Worker dry-run bundle
- `git diff --check origin/main...HEAD`

No changeset, public docs, starter-template update, or screenshots are needed because this changes only internal PartyKit persistence timing and telemetry.


spencerc99:
## Code review

PR #442, opened 2026-08-29
wewere.online: live departures board from navigation events

spencerc99:
## Summary

Adds a live train-station departures board to wewere.online at `/departures/`, rendering navigation events as departures: each row is a traveler (color dot) hopping to a new domain, with favicon, origin domain, dwell time on the previous site, local time, and a live status.

- **Shared component** `extension/website/shared/components/Departures.tsx` + SCSS — Martian Mono board text, Lora header, design-system palette (linen/paper/teal/rust), split-flap roll-in animation for new rows, scrolling notice bar, live clock.
- **Row derivation** (`deriveDepartures`): one row per domain hop per participant; collapses to the newest row per traveler+domain so ping-ponging doesn't fill the board; clamps client-clock-skewed "future" timestamps so new arrivals always sort to the top; dwell ("spent") measured from arrival on the previous domain, capped at 4h.
- **Status**: `departing` (teal flicker, hop < 2 min old) → `arrived` (rust, traveler still there) → `departed` (muted, traveler moved on). A ticking clock re-evaluates every 15s.
- **Live stream**: the worker's `/stream` WebSocket now carries navigation events alongside cursor events, with a per-socket `?types=` filter (`LiveEventsHub` filters replay + broadcast per connection). Clients that don't pass `types` get cursor-only, so the deployed portrait page is unaffected. `useLiveEvents` gained a `types` option. The departures page seeds from `/events/recent` and then flips in live rows from the socket (60s poll kept as fallback).

## Verification

- `bun run test:extension-worker`: 134 tests pass, including new coverage for per-socket type filtering (subscribed types only, filtered replay, invalid `types` falls back to cursor-only) and cursor+navigation broadcast forwarding. Two existing tests asserting navigation events were dropped from the stream were updated to the new intended behavior.
- `bun run check:extension-website` and worker typecheck clean.
- Worker deployed; verified against production: a `?types=navigation` socket received real navigation frames (URLs + participant colors enriched), a default socket still received cursor-only.
- Board verified live in-browser against real data (desktop + mobile); screenshots in the PR Evidence comment.

## Notes

- No changeset: no `packages/` changes (website + worker only, which deploy independently). No `extension/PENDING.md` bullet per the website/worker rule.
- Navigation domains were already public via `/events/recent`; the stream exposes the same data, no new surface.

### DIFF
--- a/extension/website/departures/departures.scss
+++ b/extension/website/departures/departures.scss
@@ -1,9 +1,9 @@
 // ABOUTME: Page layout for the departures board page.
-// ABOUTME: Centers the board on a dark station-hall background.
+// ABOUTME: Centers the board on the warm linen site background.
 
 body {
   margin: 0;
-  background: #1a1a1d;
+  background: #faf7f2;
 }
 
 .departures-page {
--- a/extension/website/departures/departures.scss
+++ b/extension/website/departures/departures.scss
@@ -1,0 +1,22 @@
+// ABOUTME: Page layout for the departures board page.
+// ABOUTME: Centers the board on a dark station-hall background.
+
+body {
+  margin: 0;
+  background: #1a1a1d;
+}
+
+.departures-page {
+  min-height: 100vh;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  padding: 2rem 1rem;
+  box-sizing: border-box;
+}
+
+.departures-error {
+  color: #c4724e;
+  font-family: monospace;
+}
--- a/extension/website/departures/departures.tsx
+++ b/extension/website/departures/departures.tsx
@@ -1,0 +1,48 @@
+// ABOUTME: Entry point for the departures board page on wewere.online.
+// ABOUTME: Fetches recent navigation events and renders them as a station board.
+
+import "./departures.scss";
+import React, { useCallback, useEffect, useState } from "react";
+import ReactDOM from "react-dom/client";
+import { Departures } from "../shared/components/Departures";
+import { RECENT_EVENTS_URL } from "../shared/config";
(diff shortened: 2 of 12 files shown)


spencerc99:
## Preview evidence

**Live departures board at /departures/: navigation events as train departures with color, favicon, origin, dwell, and live status.**

[Open preview](http://localhost:5194/departures/) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-mtdxow5o-c0d37b44) · Commit `8ea9634`

### Departures board (desktop)

Departures board (desktop)

Real data, 20 rows: all three statuses, varied dwell (43m), favicons. Times are UTC from headless capture.

[Open captured page](http://localhost:5194/departures/) · 1280×1000

### Departures board (mobile)

Departures board (mobile)

375px: from/spent columns drop so destination, time, status stay readable.

[Open captured page](http://localhost:5194/departures/) · 375×812

spencerc99:
## Code review

- **[P2] Keep the clock-skew clamp stable before ordering and truncation** — `extension/website/shared/components/Departures.tsx:47-49,57-59,73-79,101-107`. The code orders and selects rows using the raw client timestamp, then replaces a future timestamp with a fresh `Date.now()` every time `deriveDepartures` runs. Because the live stream and fallback poll repeatedly change `events`, the same future-dated event is reclamped to the new current time on every update: it stays `departing`, sorts above genuine new rows, and can keep consuming one of the 20 slots indefinitely—the exact pinning behavior this clamp is meant to prevent. Normalize each event to a stable effective timestamp when it first enters client state (keyed by event id), and use that same value for ordering, participant arrival/dwell state, max-row selection, display, and freshness.

PR #307, opened 2026-07-13
extension: prove ownership for participant color updates

spencerc99:
## Summary
- require a P-256 proof of possession for participant color PUT requests
- bind each proof to the participant ID, color, and monotonic stored version
- reject stale replayed updates atomically in the database

## Rationale
A public `pk_` identifier is a public key, not authorization. The worker now verifies the corresponding private-key signature before accepting a color update.

## Verification
- `bun run test` in `extension/worker` (46 tests)
- `bun run test --run` in `extension` (339 tests)
- `bun run test` in `packages/playhtml` (361 tests)
- `bun run test` in `packages/react` (43 tests)
- `bun run build-packages`
- `bunx wxt build` in `extension`
- `bun run deploy --dry-run` in `extension/worker`

## PR #264 overlap
#264 is still open and adds a broader PlayHTML auth protocol. This PR intentionally uses the existing extension ECDSA identity storage and a worker-local request format, without importing #264's unreleased common auth contract. The only conceptual overlap is shared P-256 identity usage; the changed files do not overlap #264's extension files (`background.ts`, `content.ts`).

No `extension/PENDING.md` entry: this is a transparent security hardening change, not user-facing shipped behavior.

PR #443, opened 2026-08-30
Fix 6 issues from codebase audit: privacy, auth, and correctness bugs

spencerc99:
## Summary

Fixes findings #2–#8 from a scheduled read-only codebase audit (finding #1, a PartyKit shared-permission overwrite bug, was intentionally left out — it may overlap with the in-flight identity/permissions PR #264 and needs that diff checked first before a fix is written). Six independent fixes, one per commit:

- **`extension`: strip query strings/fragments from uploaded event URLs** — `meta.url` was uploaded to the shared worker unmodified for *every* event type (cursor, viewport, keyboard — not just navigation), leaking search queries, tokens, and other sensitive query-string content through a publicly-served endpoint (`GET /events/recent`). Local IndexedDB storage keeps the full URL for the user's own history features; only the outgoing payload is sanitized.
- **`extension`: require signed ownership proof for quarantine-tape writes** — `createdBy`/`by` on the strip/rip endpoints were taken verbatim from the request body with no proof of identity, letting anyone forge another participant's identity to burn their rip allowance or frame their identity in the public verdict feed (same problem class as the open participant-color-proof PR #307, different endpoint). Payloads are now signed with the caller's ECDSA identity and verified against their public-key participant id.
- **`@playhtml/react`: fix function-form `defaultData` crash** — `(el) => ({ text: el.id })`-style `defaultData`/`myDefaultAwareness` was resolved against `ref.current` during the initial render, before React attaches the ref, crashing any consumer using the documented pattern. Resolution now defers to a layout effect.
- **`@playhtml/react`: fix stale `usePresence().myIdentity`** — it was memoized once at sync completion and never updated afterward, so a later identity change (e.g. the "we were online" extension injecting identity post-sync) went unseen. It now subscribes to identity changes like `usePlayerIdentity` does.
- **`@playhtml/common`: fix `can-grow` document-listener leak** — `onMount` added document-level `keydown`/`keyup` listeners on `mouseenter` but only removed them on `mouseleave`. Removing the element while hovered (SPA navigation, React unmount) skips `mouseleave`, leaking the listeners for the life of the page.
- **`partykit`: fail closed on missing `ADMIN_TOKEN`** — a missing/unset admin secret silently authorized every admin request (hard-reset, raw-snapshot-restore) instead of rejecting. Now rejects with 500.
- **`partykit`: stop silently dropping a due autosave retry** — `onAlarm` cleared the save-retry marker before checking whether the room was actually ready to persist; if a concurrent admin operation held the write lock (or persistence was transient) when the retry fired, the save was dropped with no reschedule. The marker is now only cleared once the save is attempted.

## Verification

- `bun run check:common`, `check:playhtml`, `check:react`, `check:extension`, `check:extension-worker`, `check:partykit` — all clean
- `bun run test:common` (18/18), `test:playhtml` (535/535), `test:react` (56 + 3 integration), `test:extension` (909/909), `test:extension-worker` (138/138), `bun test` in `partykit/` (297/297) — all green
- `bun run build-packages` — all four published packages (`common`, `extension-types`, `playhtml`, `react`) build cleanly
- Every fix has a dedicated regression test; for the `can-grow` leak and the PartyKit autosave-retry fix, I verified each new test actually fails against the pre-fix code (temporarily reverted, confirmed red, restored) before finalizing
- Added changesets for `@playhtml/common` and `@playhtml/react` (patch); `playhtml` and `@playhtml/extension` pick up the `@playhtml/common` bump automatically via `updateInternalDependencies`. No changeset needed for the `extension`/`extension/worker`/`partykit` changes — none are published npm packages.
- No `apps/docs/` changes: none of these fixes change documented public API/behavior — the React fixes make the implementation match already-correct docs, and the rest are internal hardening.
- No `extension/PENDING.md` bullet: both extension-side fixes are invisible backend/privacy hardening with no user-visible workflow change to describe in plain language.

Not applicable — no new extension feature is added. The quarantine-tape and URL-upload changes modify existing, already-shipped functionality under test coverage (unit + integration tests listed above), not a new user-facing capability.

---

### DIFF
--- a/.changeset/fix-can-grow-listener-leak.md
+++ b/.changeset/fix-can-grow-listener-leak.md
@@ -1,0 +1,5 @@
+---
+"@playhtml/common": patch
+---
+
+Fix a `can-grow` element leaking document-level keydown/keyup listeners when it's removed from the DOM while hovered (e.g. an SPA route change or React unmount that doesn't fire `mouseleave` first). The listeners are now cleaned up on unmount regardless of hover state.
--- a/.changeset/fix-react-identity-resolution.md
+++ b/.changeset/fix-react-identity-resolution.md
@@ -1,0 +1,5 @@
+---
+"@playhtml/react": patch
+---
+
+Fix two bugs in `@playhtml/react`: function-form `defaultData`/`myDefaultAwareness` (e.g. `(el) => ({ text: el.id })`) was resolved against `null` during the first render instead of the real element, crashing on mount — it now resolves once the element is attached. Also, `usePresence().myIdentity` used to freeze at sync completion and never reflect a later identity change (e.g. the "we were online" extension injecting identity post-sync); it now updates reactively like `usePlayerIdentity`.
--- a/extension/src/__tests__/bottles.visibility.test.ts
+++ b/extension/src/__tests__/bottles.visibility.test.ts
@@ -67,6 +67,7 @@ describe("bottles inventory visibility", () => {
       presence: {} as PresenceAPI,
       playerColor: "#4a9a8a",
       playerPid: "me",
+      signPlayerPayload: async () => null,
       inventory: inventoryManager.api,
     });
     const host = document.querySelector<HTMLElement>("#we-were-online-bottles");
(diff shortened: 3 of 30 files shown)


spencerc99:
Splitting this into separate PRs per fix, and folding the quarantine-tape ownership-proof fix into #307 (same problem class, same underlying signing mechanism). Closing this one — see individual PRs to follow.

---

PR #445, opened 2026-08-31
extension: strip query strings/fragments from uploaded event URLs

spencerc99:
## Summary

`meta.url` was uploaded to the shared worker unmodified for *every* event type (cursor, viewport, keyboard — not just navigation), leaking search queries, tokens, and other sensitive query-string content through a publicly-served endpoint (`GET /events/recent`). The only gate on that endpoint is a client-spoofable Origin/Referer check, documented in `originAllowlist.ts` as "a casual-scraping deterrent, not a security boundary."

Fix: strip the query string and hash fragment from `meta.url` right before upload in `extension/src/storage/sync.ts`, leaving the local IndexedDB copy (used for the user's own history/visualization features) untouched with the full URL.

## Verification

- `bun run check:extension` — clean
- `bun run test -- src/__tests__/sync.test.ts` — 2/2 pass (verifies query/hash stripped for a non-navigation event type, and that the original event object passed in isn't mutated)

Not applicable — no new extension feature is added; this modifies an existing upload path under unit test coverage, not a new user-facing capability.

---

### DIFF
--- a/extension/src/__tests__/sync.test.ts
+++ b/extension/src/__tests__/sync.test.ts
@@ -1,0 +1,56 @@
+// ABOUTME: Tests that event upload strips query strings/fragments from meta.url
+// ABOUTME: before events leave the device, without touching the caller's objects.
+
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { uploadEvents } from "../storage/sync";
+import type { CollectionEvent } from "../collectors/types";
+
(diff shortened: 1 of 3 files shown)


spencerc99 (commented):
## Code review

spencerc99:
hmm no but this breaks youtube and other things where the search query is needed for making it different from every other url - we could probably change it to just strip hash?>

PR #446, opened 2026-08-31
react: fix function-form defaultData crash on initial render

spencerc99:
## Summary

`packages/react/src/index.tsx` resolved function-form `defaultData`/`myDefaultAwareness` (e.g. `(el) => ({ text: el.id })` — the documented pattern shown in `apps/docs/src/content/docs/custom-elements.mdx`) synchronously during the initial render, calling the consumer's function with `ref.current` while it's still `null` (React hasn't attached the ref yet at that point). Any consumer passing the function form crashes on first render.

The vanilla core library already resolves this correctly against the real element (`packages/playhtml/src/elements.ts`), so only the React wrapper's own local `data`/`myAwareness` state was affected — not the persisted value once sync catches up.

Fix: seed React's initial state with `undefined` when `defaultData`/`myDefaultAwareness` is a function, and resolve it in a layout effect (`useElementRegistrationEffect`, already used elsewhere in this file) once `ref.current` is populated.

## Verification

- `bun run check:react` — clean
- `bun run test` in `packages/react` — 58/58 pass, including two new regression tests that render `CanPlayElement` with function-form `defaultData`/`myDefaultAwareness` and assert against the real element instead of crashing
- Added a changeset (`@playhtml/react` patch)

Found via a scheduled read-only codebase audit. A prior attempt at this exact fix exists on branch `cx/fix-react-function-defaults`, but that branch reverts its own change, so the bug is still live on `main`.

---

### DIFF
--- a/.changeset/fix-react-default-data-crash.md
+++ b/.changeset/fix-react-default-data-crash.md
@@ -1,0 +1,5 @@
+---
+"@playhtml/react": patch
+---
+
+Fix function-form `defaultData`/`myDefaultAwareness` (e.g. `(el) => ({ text: el.id })`) being resolved against `null` during the initial render instead of the real element, crashing on mount. It now resolves once the element is attached.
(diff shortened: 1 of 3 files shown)


spencerc99 (commented):
## Code review

PR #447, opened 2026-08-31
react: fix usePresence().myIdentity freezing at sync completion

spencerc99:
## Summary

`usePresence().myIdentity` (`packages/react/src/hooks.ts`) was a `useMemo` keyed only on `isLoading`, which transitions false→true exactly once — so `myIdentity` was computed the instant sync finished and never re-derived afterward. The sibling `usePlayerIdentity` hook's own docstring explicitly calls out that the "we were online" extension can inject identity post-sync via a `playhtml:configure-identity` event, and correctly re-subscribes to `playhtml.users.onChange` to catch it — `usePresence`'s `myIdentity` did not, so a component reading it could show a stale/wrong player color or pid indefinitely after such a change.

Fix: subscribe to `playhtml.users.onChange` the same way `usePlayerIdentity` does, re-reading `playhtml.presence.getMyIdentity()` on each notification with a value-equality check (dedupe) so consumers don't re-render on every presence tick.

## Verification

- `bun run check:react` — clean
- `bun run test` in `packages/react` — 57/57 pass, including a new regression test that simulates a post-sync identity change and asserts `myIdentity` updates instead of staying frozen (also extended the test mock's `getMyIdentity` to return a well-formed `PlayerIdentity` shape with a `setMockPlayerIdentity` helper for tests to trigger a change)
- Added a changeset (`@playhtml/react` patch)

---

### DIFF
--- a/.changeset/fix-use-presence-identity.md
+++ b/.changeset/fix-use-presence-identity.md
@@ -1,0 +1,5 @@
+---
+"@playhtml/react": patch
+---
+
+Fix `usePresence().myIdentity` freezing at sync completion and never reflecting a later identity change (e.g. the "we were online" extension injecting identity post-sync). It now updates reactively, mirroring `usePlayerIdentity`.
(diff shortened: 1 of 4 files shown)


spencerc99 (commented):
## Code review

PR #448, opened 2026-08-31
common: fix can-grow leaking document keydown/keyup listeners

spencerc99:
## Summary

`can-grow`'s `onMount` (`packages/common/src/index.ts`) adds `document`-level `keydown`/`keyup` listeners on `mouseenter`, but only removes them on `mouseleave`. `onMount` returned no cleanup function, so `ElementHandler.destroy()` had no path to remove them. If a hovered `can-grow` element is removed from the DOM without a preceding `mouseleave` — an SPA route change, a React unmount, any dynamic removal, since `mouseleave` doesn't fire on programmatic removal — the document listeners persist for the life of the page, firing `canGrowCursorHandler` against the orphaned handler on every keystroke thereafter.

Fix: return a cleanup function from `onMount` (the type already supports this — `ElementHandler` already wires up `onUnmount` when `onMount` returns one) that removes the listeners unconditionally.

## Verification

- `bun run check:common`, `check:playhtml` — clean
- `bun run test` in `packages/playhtml` — 16/16 pass in the affected test file, including two new regression tests
- Verified the tests actually catch the regression: temporarily reverted the fix, confirmed both new tests fail (`removeEventListener` not called; a leaked `keydown` listener changes the cursor after destroy), then restored the fix and confirmed green
- Added a changeset (`@playhtml/common` patch — `playhtml` picks this up automatically via `updateInternalDependencies`)

---

### DIFF
--- a/.changeset/fix-can-grow-listener-leak.md
+++ b/.changeset/fix-can-grow-listener-leak.md
@@ -1,0 +1,5 @@
+---
+"@playhtml/common": patch
+---
+
+Fix a `can-grow` element leaking document-level keydown/keyup listeners when it's removed from the DOM while hovered (e.g. an SPA route change or React unmount that doesn't fire `mouseleave` first). The listeners are now cleaned up on unmount regardless of hover state.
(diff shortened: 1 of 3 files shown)


spencerc99 (commented):
## Code review

PR #449, opened 2026-08-31
partykit: stop silently dropping a due autosave retry

spencerc99:
## Summary

`onAlarm`'s document-save retry (`partykit/party.ts`) cleared its retry marker (`clearDocumentSaveRetry()`) *before* checking whether the room was actually `roomState() === "ready"`. If a concurrent admin operation held the write lock (`documentMaintenanceInProgress`, e.g. `restoreFromSnapshot`/hard-reset in flight) or persistence was transient (Supabase unreachable) when the retry fired, the save attempt was skipped entirely — but the marker was already gone, so the save was silently dropped forever unless some later edit happened to re-trigger a fresh autosave cycle.

The sibling empty-room-compaction retry three lines below already gets this right: it only clears its own marker once the operation actually ran or is confirmed no longer applicable. This PR applies the same pattern to the document-save retry — the marker is now cleared only once the save is attempted; otherwise it's rescheduled for the next alarm.

## Verification

- `bun run check:partykit` — clean
- `bun test` in `partykit/` — 291/291 pass, including two new regression tests (one exercising the "mid-maintenance" case via `onAlarm()` directly, one via the platform `alarm()` entry point) plus an updated existing test that previously asserted the buggy drop-on-transient-persistence behavior as expected
- Verified the tests actually catch the regression: temporarily reverted the fix, confirmed all three fail, then restored and confirmed green

This fix is fully independent of `ADMIN_TOKEN`/admin-route auth — `onAlarm` is the platform's timer callback, not an HTTP route, so none of the new/updated tests need any admin token.

---

### DIFF
--- a/partykit/__tests__/quarantineLoad.test.ts
+++ b/partykit/__tests__/quarantineLoad.test.ts
@@ -2864,7 +2864,7 @@ describe("quarantine data safety", () => {
     expect(persistedRow.document).toBe(candidate.base64);
   });
 
-  test("an unavailable room consumes an expired save retry", async () => {
+  test("an unavailable room reschedules (does not drop) an expired save retry", async () => {
     const { room, storage } = createRoom();
     room.documentLoadCompleted = true;
     room.persistenceMode = {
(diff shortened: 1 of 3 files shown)


spencerc99 (commented):
## Code review

PR #450, opened 2026-08-31
partykit: fail closed on missing ADMIN_TOKEN

spencerc99:
## Summary

`getAdminAuthError` (`partykit/adminAuth.ts`) returned `null` (authorized) whenever `adminToken` was falsy: `if (!adminToken) return null;`. If `ADMIN_TOKEN` is ever unset for a deployment — a misconfigured secret, a new preview/staging environment spun up without it — every admin route becomes open with no authentication, including hard-reset and raw-snapshot-restore, both of which can overwrite or destroy live room data. This is a fail-*open* default rather than fail-closed.

Fix: treat a missing token as a 500 config error rather than an authorization pass.

## Verification

- `bun run check:partykit` — clean
- `bun test` in `partykit/` — 295/295 pass, including a new dedicated `adminAuth.test.ts` (6 tests covering the missing-token, empty-token, wrong-token, and correct-token-via-query-param/header cases)
- This change makes every admin-route test that previously ran with no token now need one — updated `quarantineLoad.test.ts` and `quarantinePlatformEntry.test.ts` to configure an `ADMIN_TOKEN` in their mocked env and pass a matching `?token=` on each admin request they exercise, so those tests keep asserting the actual route behavior they were written for rather than incidentally hitting a new 500

---

### DIFF
--- a/partykit/__tests__/adminAuth.test.ts
+++ b/partykit/__tests__/adminAuth.test.ts
@@ -1,0 +1,58 @@
+// ABOUTME: Tests the shared admin-token gate used by PartyKit and Worker admin routes.
+// ABOUTME: A missing ADMIN_TOKEN must fail closed, not silently authorize every request.
+import { describe, expect, it } from "bun:test";
+import { getAdminAuthError } from "../adminAuth";
+
+function request(url: string, headers: Record<string, string> = {}): Request {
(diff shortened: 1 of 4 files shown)


spencerc99 (commented):
## Code review

PR #453, opened 2026-08-31
Fix Internet Commute train assignment

spencerc99:
## Summary

- count only riders with an active boarding lease when assigning train capacity
- keep rider leases alive throughout the route and retry transient initial boarding failures
- exclude communal destinations used by retained trains when creating a route

## Why

Disconnected rider tokens stayed in dispatcher state for the entire route, so they occupied seats after their browser left. That split four live riders across two train IDs even though each UI correctly showed only its two connected riders. The dispatcher also selected the first two recent destinations for every new train, which made those split trains run identical routes.

## Verification

- `bun run test:extension-worker` — 18 files passed
- `bun run test:extension` — 140 files, 903 tests passed
- `bun run check:extension-worker`
- `bun run check:extension`
- `bun run check:extension-website`
- `WRANGLER_LOG_PATH=/private/tmp/commute-train-wrangler.log bun run smoke:extension-worker`
- automated real-browser flow on the actual Internet Commute page with the production dispatcher and route-selection modules:
  - four simultaneous sessions joined one train
  - after two sessions closed and their leases expired, two replacements reused the open seats on that train
  - a fifth active session created a second train whose two communal stops differed from the first train

Local PlayHTML cursor-presence transport could not connect because this checkout has no PartyKit/Supabase development secrets; the browser verification covers the changed dispatcher admission and route behavior directly.

No package behavior, public core API, manifest permissions, or starter-template usage changed.

### DIFF
--- a/extension/PENDING.md
+++ b/extension/PENDING.md
@@ -3,6 +3,7 @@
 - Slow Mode commute rides now connect reliably when the hosted route loads before the extension.
 - Internet Commute now stops counting riders after they leave and always shows the train pull-up and cursor boarding intro for Slow Mode rides.
 - Internet Commute now starts a fresh route after returning to home station.
+- Internet Commute now keeps active riders together and gives each new train different stops.
 
 <!--
 Add a bullet here in any PR that touches extension/**. The release-prep workflow
--- a/extension/src/entrypoints/commute/commute.tsx
+++ b/extension/src/entrypoints/commute/commute.tsx
@@ -305,6 +305,7 @@ function useCommuteTrain(
           refreshTimer = window.setTimeout(() => void board(), 3_000);
         } else {
           setConnection((current) => ({ ...current, status: "error" }));
+          refreshTimer = window.setTimeout(() => void board(), 3_000);
         }
       }
     };
--- a/extension/src/entrypoints/commute/commuteTrain.test.ts
+++ b/extension/src/entrypoints/commute/commuteTrain.test.ts
@@ -60,20 +60,25 @@ describe("commute train client", () => {
     expect(getCommuteRiderToken(null)).toBe(nextToken);
   });
 
-  it("refreshes open trains and reboards riders after completion", () => {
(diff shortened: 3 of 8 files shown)


spencerc99:
## Preview evidence

**Internet Commute keeps four concurrent riders on one train and chooses a different route for the next train without a stale-selection race.**

[Open preview](http://127.0.0.1:4174/commute/) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-mti8ob2f-42080d86) · Commit `09d5b51`

### Four concurrent riders share one route

Four concurrent riders share one route

Four simultaneous sessions joined browser-train-1 with first-one.example and first-two.example.

[Open captured page](http://127.0.0.1:4174/commute/) · 2048×1156

### Concurrent overflow uses a different route

Concurrent overflow uses a different route

The fifth simultaneous session created browser-train-2 with second-one.example and second-two.example after route selection was revalidated.

[Open captured page](http://127.0.0.1:4174/commute/) · 2048×1156

spencerc99:
## Code review

- **P1 — Revalidate used domains after the asynchronous route load** ([`extension/worker/src/commuteTrainDispatcherObject.ts:107-116`](https://github.com/spencerc99/playhtml/blob/e4b612386c94c8cf0ac358da6fcb1443c4456835/extension/worker/src/commuteTrainDispatcherObject.ts#L107-L116)). `getRecentCommunalDomains(now)` is captured before `await loadCommunalStops(...)`, and Durable Object requests can interleave while that non-storage I/O is pending. If more than four new riders board concurrently while existing trains are full, every request can select the same unseen domains; the first four fill the newly created train, then a later resumed request sees it full and creates another train from its stale, identical `communalStops`. That directly reproduces the duplicate-route behavior this diff is intended to prevent. Load the candidate destinations first, then synchronously reselect them against the dispatcher's current communal-domain set immediately before `board` (with no intervening `await`), or otherwise serialize/version-and-retry this selection-and-board critical section.

PR #456, opened 2026-09-01
Prevent options page storage stats crash

spencerc99:
## Summary

- tolerate storage stats responses without per-type size estimates
- keep rendering the available totals while omitting the Internet Scraps size
- cover the previous background worker response shape with a regression test

## Root cause

The options page could load updated UI code while the previous MV3 background worker was still active. That worker returned storage stats without estimatedSizeBytesByType, and the options page read .element from the missing map during render.

## Verification

- bun run test:extension (144 files, 915 tests)
- bun run check:extension
- bun run build-extension
- bun run smoke:extension
- loaded the built MV3 options page in isolated Chromium with the previous stats response shape; Your data rendered with zero uncaught errors

This is a narrow extension UI fix. It does not change permissions, collection behavior, or public release notes.

### DIFF
--- a/extension/src/__tests__/Collections.test.tsx
+++ b/extension/src/__tests__/Collections.test.tsx
@@ -131,6 +131,36 @@ describe("Collections", () => {
     }
   });
 
+  it("renders storage stats from a background worker without per-type sizes", async () => {
+    vi.mocked(browser.runtime.sendMessage).mockImplementation(async (message) => {
+      if (message?.type === "GET_STORAGE_STATS") {
+        return {
+          success: true,
+          stats: {
+            totalEvents: 12,
+            estimatedSizeBytes: 1536,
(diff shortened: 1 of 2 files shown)


spencerc99:
## Preview evidence

**The built extension options page renders storage totals when the active background worker does not provide per-type size estimates.**

[Open preview](https://github.com/spencerc99/playhtml/pull/456) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-mtj1favi-6c99344f) · Commit `ff6cb5b`

### Options page with previous stats response

Options page with previous stats response

The built MV3 options page renders Your data without errors and omits only the unavailable Internet Scraps size.

1440×900

spencerc99 (commented):
## Code review

PR #458, opened 2026-09-03
Internet map: ASCII city map of WWO browsing on wewere.online

spencerc99:
## Summary
Vendors [jackyzha0/internet-map](https://github.com/jackyzha0/internet-map) (built by Jacky from the WWO production export) into the site at `/internet-map/`, restyled to fit wewere.online while keeping the baked-ASCII-grid design intact.

- **Theme system** (`src/theme.ts`): `dark` is Jacky's original untouched; `paper`, `ink`, `atlas` are light presets built from the WWO token set. On light ground the prominence signal inverts (dark/saturated = busy).
- **Theme dev panel** (`?dev=1` or backtick): every color slot editable, sliders for alphas/label size/far-zoom contrast, copy/paste theme as JSON, share link, PNG export.
- **Domain search** (`/` or Cmd+K, `?q=`): ranked domain/subdomain search, flies the camera to the district and marks the landing.
- **Label anchor fix**: place names anchor on the densest cluster of their own buildings instead of the arithmetic centroid, which floated labels over empty ground for split domains.
- Site build target raised to es2022 (upstream uses top-level await).

## Data and privacy
**No map data is committed.** Bundles carry real participants' browsing labels; `extension/website/public/internet-map/data/` is gitignored and the deployed page will 404 on data until we pick a hosting mechanism for bundles (discussion in progress; default build keeps full search queries per upstream's redaction policy, so the full-map bundle is NOT public-ready as-is). `pii_audit.py` passes CLEAN on the current full build.

MEK typefaces are vendored per Jacky's usage; license confirmation from the foundry is pending before wide publicity.

## Testing
- `bun run check:extension-website` and full `bun run lint` pass.
- Verified headless (Playwright) against the 5% sample, the full 12.4M-visit build, and a personal build: presets, dev panel round-trip, search fly-to on data-full, label fix on freesewing.eu.
- Screenshots attached in comments.


spencerc99 (commented):
## Code review

### Medium

- `extension/website/internet-map/src/ascii/renderer.ts:27` — The cache budget permits a 64-million-pixel offscreen canvas. On the initial zoomed-out view, `draw()` selects the cache for maps over 520k visible cells, so the full 4096-column map can eagerly allocate roughly 256 MB just for the RGBA canvas (before the baked-map arrays and browser overhead). Mobile browsers and lower-memory devices can reject that canvas or kill the page, and the non-null `getContext("2d")!` then offers no recovery. Use a substantially smaller/device-aware pixel budget, cap each canvas dimension to supported limits, and fall back to live/tiled rendering when context creation fails.

- `extension/website/internet-map/index.html:28` — This adds a new user-facing interactive page, but the PR conversation contains no screenshots even though the PR says they are attached. The applicable repository instructions require screenshots from the real affected surface for user-facing changes, including key interaction states. Attach captures of the actual map page covering the initial view and the important search/routing/theme-panel states so the finished UI can be reviewed.

PR #459, opened 2026-09-05
Avoid redundant document connections for domain-wide cursors

spencerc99:
Domain-wide cursors (`cursors: { enabled: true, room: "domain" }`) created a separate Yjs document provider even when the dedicated presence transport already handled cursor traffic. A page on spencer.place therefore opened an unnecessary `/parties/main/spencer.place` connection alongside its page document and presence connections.

Resolve the cursor room and acquire its presence transport without creating a separate document provider. Remove the unused cursor-provider cleanup and host/error arguments. The missing-WebSocket condition cannot provide a working document-provider fallback because the main Yjs provider also requires WebSocket.

The September 2 traffic investigation found 385,141 requests to the domain document path versus 1,203 across page document paths. This change removes that unnecessary client connection; the exact cause of repeated reconnects was not reproduced, and the production traffic reduction must be measured after rollout.

Validation:

- All 537 core tests passed on 4a918b67, including domain and custom cursor-room coverage verifying that only the main document provider is created.
- Core type check and core package build passed on 4a918b67; all package builds passed on the preceding commit.
- A real Chromium browser on the local domain-cursor playground attempted one page document path and two presence paths (page and domain), with no domain document path. The local backend was unavailable, so this verifies connection selection, not successful end-to-end synchronization.

Includes a patch changeset. No public API, setup, or persisted-data changes; public docs and starter templates need no changes. Production rollout requires a package release and upgrading/redeploying spencer.place.

### DIFF
--- a/packages/playhtml/src/index.ts
+++ b/packages/playhtml/src/index.ts
@@ -1218,29 +1218,30 @@ function buildCursors(args: {
 
   if (cursorOptions.room) {
     const cursorRoomString = resolveCursorRoom(cursorOptions.room);
-    const cursorRoom = normalizeRoomId(getCurrentRoomHost(), cursorRoomString);
-
-    if (cursorRoom !== mainRoom) {
-      const cursorDoc = new Y.Doc();
-      cursorProvider = new YProvider(
-        partykitHost,
-        cursorRoom,
-        cursorDoc,
-      );
-      cursorProvider.on("error", () => {
-        onError?.();
-      });
-      providerForCursors = cursorProvider;
-      currentCursorRoomId = cursorRoom;
-    } else {
-      currentCursorRoomId = mainRoom;
-    }
+    currentCursorRoomId = normalizeRoomId(
+      getCurrentRoomHost(),
+      cursorRoomString,
+    );
   } else {
     currentCursorRoomId = mainRoom;
   }
 
   const cursorPresenceTransport =
     acquirePresenceTransport(currentCursorRoomId) ?? undefined;
+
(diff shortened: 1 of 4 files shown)


spencerc99 (commented):
## Code review

packages/playhtml/src/index.ts
```diff
@@ -1218,29 +1218,30 @@ function buildCursors(args: {
...
+    cursorProvider = new YProvider(
+      partykitHost,
+      currentCursorRoomId,
+      cursorDoc,
+    );
+    cursorProvider.on("error", () => {
+      onError?.();
+    });
+    providerForCursors = cursorProvider;
+  }
```
spencerc99: do we need this at this point now that we've migrated to the presence transport? When would we not have a presence transport?

spencerc99: Spencer, we do not need this separate-provider fallback. `canUseRealtimePresenceTransport()` only checks whether the global `WebSocket` exists. It does not become false when the presence server is down or a socket disconnects.

Without WebSocket, the main Yjs provider cannot connect either: y-partyserver reports "No WebSocket implementation available", and PlayHTML passes no WebSocketPolyfill. My earlier test reached the fallback by removing WebSocket while using a fake Yjs provider; it did not demonstrate a working runtime fallback.

Removed the separate cursor provider, its cleanup, and its unused host/error arguments in 4a918b67. Domain and custom cursor-room tests now verify that only the main document provider is created. All 537 core tests, the core type check, and the package build pass. The broader awareness fallback code is outside this removal.

PR #463, opened 2026-09-06
Enable sound on the installation cursor screen

spencerc99:
## Summary

- enable sound by default only for the named `screen=cursors` installation profile
- retry suspended audio when the page becomes visible or returns from the back-forward cache
- keep audio initialization available when autoplay leaves a resume request pending
- remove per-frame resume retries so blocked playback cannot accumulate pending promises

Followers and the scrolling, keypresses, touches, and clicks screens remain silent by default.

## Verification

- extension tests: 988 passed
- focused sound and profile tests: 18 passed
- extension website typecheck passed
- extension website production build passed
- cursor and follower localhost routes loaded successfully
- two independent diff reviews completed; all findings addressed

## Installation requirement

The webpage does not bypass browser autoplay policy. Configure the main cursor computer with an autoplay allowlist for `https://wewere.online` or launch its dedicated Chrome session with `--autoplay-policy=no-user-gesture-required` for reliable unattended cold starts.

### DIFF
--- a/extension/website/installation/live/live.tsx
+++ b/extension/website/installation/live/live.tsx
@@ -95,6 +95,7 @@ const LiveInstallation = () => {
         activeVisualizations={activeVisualizations}
         onSetActiveVisualizations={setActiveVisualizations}
         availableVisualizations={LIVE_INSTALLATION_VISUALIZATIONS}
+        defaultSoundEnabled={profile?.defaultSoundEnabled}
         defaultSettings={settingsDefaults}
         useStoredSettings={profile === null}
         syncSettingsToUrl={profile === null}
--- a/extension/website/shared/components/MovementCanvas.tsx
+++ b/extension/website/shared/components/MovementCanvas.tsx
@@ -14,6 +14,7 @@ import { AnimatedTrails } from "./AnimatedTrails";
 import { LiveTrails } from "./LiveTrails";
 import { LiveIndicator } from "./LiveIndicator";
 import { SoundEngine } from "../sound/SoundEngine";
+import { attachSoundWakeListeners } from "../sound/soundWake";
 import { AnimatedClicks, type ScheduledClick } from "./AnimatedClicks";
 import { AnimatedTyping } from "./AnimatedTyping";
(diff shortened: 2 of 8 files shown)


spencerc99 (commented):
## Code review

### Medium severity

- `extension/website/shared/utils/liveInstallationProfiles.ts:181` — This enables a new unattended, user-facing sound behavior, but the verification only says the cursor and follower localhost routes loaded; it does not exercise the AudioContext under the autoplay configuration this feature explicitly depends on. A route load can pass while the context remains suspended, which would leave the installation silently broken, and the repository instructions require complete real-browser end-to-end verification for new extension behavior. Run the built or deployed `?screen=cursors` profile in a real Chrome session with the production autoplay allowlist/launch flag, verify a cold start and a visibility/BFCache wake leave the context running and audible, and verify the follower plus other named profiles remain silent; record those results before merge.

spencerc99:
Real-Chrome verification for `13c25408` is complete.

- Ran the built site in installed Google Chrome with `--autoplay-policy=no-user-gesture-required`. Playwright's default `--mute-audio` and `--disable-back-forward-cache` flags were omitted.
- `screen=cursors` cold start created one `AudioContext` in `running` state. Live cursor trails started two oscillators immediately and 14 before navigation. No page errors occurred.
- A same-origin navigation to `/privacy.html` and browser Back restored the original document from the back/forward cache. The document probe ID stayed the same, `pagehide.persisted` and `pageshow.persisted` were both `true`, visibility changed from hidden to visible, the same audio context stayed running, and oscillator starts advanced from 14 to 16 after restore.
- `follower-a`, `scrolling`, `typing`, `clicks`, and `touches` each created zero audio contexts and zero oscillator starts during their checks.
- The harness wrapped `AudioContext` before page load only to observe context state, resume calls, and source starts. It did not change application behavior.

This proves that the audio graph runs and produces source activity in the intended profile, including after a real back/forward-cache restore. It does not prove physical speaker audibility or the production machines' Chrome policy. That final check still requires the actual installation hardware, output device, and launch configuration.

PR #464, opened 2026-09-07
Fix extension Worker deployment verification

spencerc99:
The extension Worker deploy token can publish Workers but cannot query D1. The installation-control deploy failed before publishing because the migration step returned Cloudflare error 7403.

This removes the unsupported migration command from the deploy workflow and verifies the live installation-control contract after every Worker deployment. The verification checks the response shape plus the required CORS and `no-store` headers.

The pending `0002_quarantine_tape_feature.sql` and `0003_installation_control.sql` migrations were applied separately with the authorized local Wrangler 4.81.1 session. The Worker was then deployed as version `7ee63be9-2e05-46f7-b04f-7ce2dd20351e`, and the production endpoint returned HTTP 200 with generation `0`, `Access-Control-Allow-Origin: *`, and `Cache-Control: no-store`.

Validation:

- `bunx prettier --check .github/workflows/extension-worker-deploy.yml`
- `git diff --check`
- The workflow's `jq` contract expression passes against the production payload.

### DIFF
--- a/.github/workflows/extension-worker-deploy.yml
+++ b/.github/workflows/extension-worker-deploy.yml
@@ -51,14 +51,6 @@ jobs:
       - name: Bundle extension Worker
         run: bun run smoke:extension-worker
 
-      - name: Apply extension Worker database migrations
-        uses: cloudflare/wrangler-action@v4
-        with:
-          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-          accountId: ${{ vars.CLOUDFLARE_ACCOUNT_ID }}
-          packageManager: bun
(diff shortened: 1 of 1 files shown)


spencerc99 (commented):
## Code review

### Medium severity

- `.github/workflows/extension-worker-deploy.yml:54` — The deployment now proceeds directly from bundling to publishing and permanently removes the repository's only remote D1 migration step. Applying migrations manually fixed the current production database, but the next Worker change that depends on a new migration can now deploy against an older schema and fail in production; the installation-control GET only proves one current table exists and cannot detect general migration drift. Keep migration application as a required pre-deploy step using a separate D1-capable token/secret (or an equivalently enforced migration workflow), and block Worker publication unless it succeeds.

PR #465, opened 2026-09-07
extension: add local installation cursor mode

spencerc99:
Installation displays can show the participant’s own colored cursor across ordinary web pages. Open extension Settings, press Cmd/Ctrl+Shift+8, and enable **Installation mode** under Identity. The setting starts off, persists across browser restarts, and remains visible while enabled.

The cursor uses the existing arrow and saved profile color in a shadow-root overlay. Color and toggle changes reach open tabs through local storage. Disabling removes the overlay and restores the page’s cursor. The setting does not join multiplayer rooms or change collection preferences. Browser pages and embedded frame contents retain their native cursor.

Validation:
- Production extension build and `bun run check:extension` passed.
- Settings and presence-policy suites passed: 20 tests.
- `node smoke-tests/installation-cursor.mjs` passed against the built MV3 extension in Chromium 147.0.7727.15. The test uses the real Settings UI, service worker, identity, storage, and content script. It covers both shortcut modifiers, opt-in, color changes, clicking, typing, navigation/back/reload, browser restart, and disabling across two open tabs. No multiplayer sockets opened. External traffic was blocked; no backend responses or extension APIs were mocked.
- Finished settings and cursor screenshots are attached in PR Evidence.

This hidden installation setting has no public release-note entry. No package API, manifest permissions, or starter templates change.

### DIFF
--- a/extension/CLAUDE.md
+++ b/extension/CLAUDE.md
@@ -310,6 +310,23 @@ Cloudflare Worker + Supabase PostgreSQL + Resend:
 
 ## Configuration
 
+### Installation cursor
+
+In extension Settings, press Cmd/Ctrl+Shift+8 to reveal **Installation mode**
+under Identity. The checkbox shows the participant's own colored cursor on web
+pages and persists across browser restarts. It starts off; while enabled, its
+checkbox stays visible in Settings. Color changes apply to open pages.
+
+This is a local display setting, independent of feature entitlements and
+collection modes. It does not start cursor presence connections. Browser pages
+and embedded frame contents retain their native cursor.
+
+Build with `bun run build-extension`, then run
+`node smoke-tests/installation-cursor.mjs` to exercise the real Settings flow in
+isolated Chromium. Set `INSTALLATION_EVIDENCE_DIR` to save screenshots.
+
+### Files
+
 - `src/config.ts`: `VERBOSE` debug logging flag
 - `src/flags.ts`: Feature flags (`COPRESENCE: true`)
 - `wxt.config.ts`: Manifest v3, permissions (storage, tabs, http/https host access), React module, ASCII charset output for Chrome compliance
--- a/extension/CLAUDE.md
+++ b/extension/CLAUDE.md
@@ -319,7 +319,9 @@ checkbox stays visible in Settings. Color changes apply to open pages.
 
 This is a local display setting, independent of feature entitlements and
 collection modes. It does not start cursor presence connections. Browser pages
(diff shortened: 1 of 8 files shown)


spencerc99:
## Preview evidence

**Verified installation cursor across light DOM and open shadow roots with closed-root fallback**

[Open preview](http://127.0.0.1/) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-mtrjdw0p-04586527) · Commit `bc93813`

### Installation mode enabled

Installation mode enabled

The hidden Settings control enables the local cursor.

1100×760

### Rootless custom element

Rootless custom element

The installation cursor stays visible over a light-DOM custom element; the native pointer is suppressed and restored when disabled.

1100×760

### Open shadow root

Open shadow root

Native cursor suppression still works in open and nested late-attached roots.

1100×760

### Closed shadow root fallback

Closed shadow root fallback

The overlay hides over a closed-root button while its native crosshair remains.

1100×760

### Mode disabled

Mode disabled

Disabling restores native cursor styles across open tabs.

1100×760

spencerc99 (commented):
## Code review

### Medium severity

- `extension/src/entrypoints/content/style.css:6` — The descendant selector cannot cross a shadow-root boundary, but `pointermove` from a Web Component still reaches `initInstallationCursor`. If the component's shadow content specifies its own `cursor` (for example `pointer` on a button), the browser therefore keeps drawing that native cursor while this feature also draws the installation overlay, producing two cursors on common componentized pages. The committed smoke page only uses light-DOM controls, so it does not cover this path. Suppress the native cursor inside reachable shadow roots (including roots attached after initialization), define a safe fallback for closed roots, and add a real-browser case with a custom element that sets an internal cursor.

spencerc99:
## Code review

### Medium severity

- `extension/src/entrypoints/content/installationCursor.ts:46` — The closed-root fallback also hides the installation cursor for every light-DOM custom element that has no shadow root. In Chrome, `chrome.dom.openOrClosedShadowRoot(node)` returns `null` when there is no root, but the `!root && node.localName.includes("-")` branch still calls `hide()`. A realistic `<cursor-probe>` (or any autonomous custom element) that renders children in light DOM therefore loses the overlay and restores its native cursor even though the document-level rule can suppress it safely. Distinguish “the browser cannot inspect closed roots” from “inspection succeeded and there is no root”; only use the conservative custom-element fallback in the former case, and cover a rootless light-DOM custom element in the browser smoke.

spencerc99:
## Code review

PR #468, opened 2026-09-08
Reduce cursor trail and ripple rendering work

spencerc99:
Cursor playback was formatting every quadratic curve of every moving freehand trail into an SVG string on every frame. Reuse curve strings when their endpoint coordinates match the preceding outline, including both sides of a growing stroke. The generated SVG remains byte-for-byte identical. Cache entries use weak trail-point keys and are released when a trail completes.

Click and hold ripples now update SVG ring radii directly instead of committing React state every frame. Equal settings objects no longer cause unchanged ripples to render when other clicks arrive. Expansion math, ring delays, hold scaling, and residue are preserved. Completion callbacks are guarded across Strict Mode effect restarts.

Also use binary search for the expired-trail boundary, so each frame no longer scans all previously evicted trails.

Validation:
- Ripple benchmark: real Chromium at 1920x1080, 4x CPU throttling, 48 generated six-ring hold effects, four alternating 5-second runs. JavaScript time dropped from 3.53–3.59 seconds to 0.45–0.49 seconds (about 87% less); main-thread task time dropped from 4.97–5.00 seconds to 2.26–2.33 seconds (about 54% less). React profiler commits dropped from 446–459 to the initial mount only. No page errors.
- Full-component Chromium benchmark at 1920x1080 with 4x CPU throttling: 24 identical generated 1,200-point cursor trajectories rendered by the real `AnimatedTrails` component, using the original versus optimized helper. Four alternating runs (original/optimized/optimized/original), 5 seconds each after warm-up: original 131/139 frames (~26–28 FPS), optimized 202/212 (~40–42 FPS). Frame-time p95 improved from 50.4–51.1 ms to 33.3–33.9 ms. No page errors. This controlled desktop stress test does not emulate a specific Raspberry Pi or GPU.
- Chromium comparison against the original implementation at e45ad14d: 600 identical moving-head frames over a 1,000-point curved trajectory, five alternating-order runs. Median path-building time fell from 168.4 ms to 59.7 ms (2.8x faster, 65% less time). Additional comparison of 100 frames produced identical SVG strings. This is a geometry microbenchmark on a Mac, not a Raspberry Pi FPS measurement.
- 42 focused tests passed, covering geometry, trail renderers, live/archive playback, and cursor hooks. The no-frame-commit regression test was also run against the original ripple component and failed at the React commit assertion. Added coverage for ripple geometry, settings changes, hold lifetimes, cleanup, Strict Mode completion, growing/shrinking outlines, width changes, moving tails, completion, replay wrap, and eviction boundaries.
- Extension and website typechecks passed. Website production build passed with Vite's existing large-chunk advisory.
- Automated headless Chromium at 1920x1080 against the production-built `/installation/live/?screen=cursors` , `?screen=follower-a`, and `?screen=clicks`, using the real production event feed. Asserted changing path geometry, cursor pause/resume, follower viewBox presence, and movement after reload. No page or console errors. Screenshots cover the cursor field, frozen field, follower close-up, and click ripples. PR evidence is refreshed for the current commit.

No package API or dependency changes; changesets, starter templates, and public API docs do not apply. This is website cursor rendering maintenance and does not need a released-extension changelog entry. Raspberry Pi hardware, GPU acceleration, and the installation's exact browser configuration remain unverified.

Current commit 6bafa897 passed PR validation CI: common/core/docs/React/extension/Worker tests, lint, site build, and extension/site smoke tests.

### DIFF
--- a/extension/website/shared/components/ClickRipple.tsx
+++ b/extension/website/shared/components/ClickRipple.tsx
@@ -1,6 +1,6 @@
 // ABOUTME: Shared ripple effect for click/hold visualization
 // ABOUTME: Used by AnimatedTrails and AnimatedClicks so ripple logic stays DRY
-import { useState, useEffect, useRef, memo } from "react";
+import { useEffect, useRef, memo } from "react";
 import { ClickEffect } from "../types";
 
 export interface RippleSettings {
@@ -63,10 +63,11 @@ export const RippleEffect = memo(
     settings: RippleSettings;
     onComplete?: (id: string) => void;
   }) => {
-    const [now, setNow] = useState(Date.now());
-    const [isAnimating, setIsAnimating] = useState(true);
-    /** Ensures onComplete runs once — render-phase callbacks can run twice in Strict Mode. */
+    const now = Date.now();
+    const ringRefs = useRef<Array<SVGCircleElement | null>>([]);
+    /** Ensures onComplete runs once even when effects restart in Strict Mode. */
     const completionFiredRef = useRef(false);
+    const completionIdRef = useRef(effect.id);
 
     const lifecycle = getRippleLifecycle(effect, rippleSettings, now);
     const { holdMultiplier, expansionDuration } = lifecycle;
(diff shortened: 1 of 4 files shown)


spencerc99 (commented):
## Code review

spencerc99:
## Preview evidence

**Cursor trail and ripple optimizations retain moving geometry, frozen mode, and follower framing in the production build.**

[Open preview](http://127.0.0.1:5179/installation/live/?screen=cursors) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-mtsfzft5-3638504a) · Commit `6bafa89`

### Cursor field in motion

Cursor field in motion

Final production build; cursor paths animate and continue after reload.

[Open captured page](http://127.0.0.1:5179/installation/live/?screen=cursors) · 1920×1080

### Frozen trail field

Frozen trail field

Existing frozen mode retains stable completed geometry; resume was verified.

[Open captured page](http://127.0.0.1:5179/installation/live/?screen=cursors) · 1920×1080

### Follower camera close-up

Follower camera close-up

Final production follower profile retains camera framing and movement after reload.

[Open captured page](http://127.0.0.1:5179/installation/live/?screen=follower-a) · 1920×1080

### Expanding click rings

Expanding click rings

Final production click screen with directly animated SVG radii.

[Open captured page](http://127.0.0.1:5179/installation/live/?screen=clicks) · 1920×1080

### Accumulated click rings

Accumulated click rings

Ring expansion and accumulated marks retain their appearance.

[Open captured page](http://127.0.0.1:5179/installation/live/?screen=clicks) · 1920×1080

PR #473, opened 2026-09-08
Give the spotlight a lifecycle, and a lab to tune it in

spencerc99:
## Summary

The spotlight was a boolean — a trail either held it or it did not — and every part of a promotion had to be inferred from the edge where that boolean changed. It is now a sequence of named phases:

```
idle -> auditioning -> entrance -> reign -> release -> cooldown
```

The outlier test still decides *who*. The phases decide *when*, and each owns the part of the sound belonging to it. Alongside it: a lab on `/sounds/` for choosing the numbers by ear, and two fixes to the click scanner's mid-scene action harness.

## The lifecycle

The existing hysteresis is absorbed rather than removed. The promote ratio and floor make a candidate, the demote ratio and floor end a reign, and `minSoloDurationMs` becomes the reign's own `minReignMs`. What the phases add is that none of those tests are asked where their answer would be noise:

- a candidate has to keep clearing the promote test for a whole `auditionMs` (500ms) before it counts, which no flick of the wrist does;
- the demote test is not consulted at all until the reign has run its `minReignMs` (3s) — the one exception being a trail that leaves the scene, which is not a judgement about speed and cannot wait;
- `sceneCooldownMs` (4s) after a release nobody auditions, so a scene with two fast trails sings as a crowd for a while instead of handing the spotlight back and forth;
- nothing is promoted in a scene thinner than `minTrailsForSpotlight` (3), because a soloist is a voice stepping out of a crowd.

`entranceMs` (200ms) now drives the gain lift, the drying and the brightening together through one helper, so a promotion arrives as a single move rather than three that overlap. The vibrato is held back out of it and widened across `reignDevelopMs` (1.5s) during the reign instead, alongside the halo's own 800ms onset: the voice arrives, then it warms. `PRESENCE_TUNING.swellSeconds` is gone, replaced by the entrance everywhere it was read.

An engine carries its own copy of the values worth turning by ear; `setSpotlightTuning` moves them while a scene plays.

## The lab

Three deterministic ~20s scenarios on `/sounds/`, sliders that reach into the running engine, and a readout of which phase you are hearing. Each scene is a question the tuning has to answer:

- **one fast among five** — the case the spotlight is for. Its lead changes halfway through, so a release and a second promotion are both in earshot inside one run.
- **churny crowd** — six cursors trading the lead three times a second, faster than an audition. Nobody should be promoted.
- **single cursor** — sweeping alone, and it must never promote.

A ring marks the soloist on both the lab canvas and the replay: solid is promoted, dashed is auditioning and may come to nothing.

## The scanner harness

Two real bugs in the mid-scene action harness, both found through the no-op control scene:

1. Suspension is granted on a render quantum boundary, **rounding up**, and the harness grouped events by the quantum below — handing the engine a clock a quantum behind the render, so every ramp it anchored there began from a value the output had already passed.
2. The clock is now the render head read from the context rather than computed from the suspension request. A driven context shadows `currentTime`, so the getter is taken off the prototype chain before the shadow goes on.

Beyond that: while a tick is driven ahead of the head, each *un-timed* effect is redirected to the clock the tick is running on. A `disconnect` is queued and applied at the next suspension; a node's gain is pinned to zero between the head and the moment it was built; a `value` write is scheduled at the driven clock. Without this, a formant bank attached thirty seconds into a scene arrived at full level at the action's head, on top of a voice already sounding.

## Verification

- **Extension tests:** 1204 passing (154 files). Includes a new lifecycle suite driven on a clock the test owns — every assertion is about *when* — plus tuning-overlay roundtrip tests. The existing spotlight tests keep their structure; what changed is that a promotion now takes longer than a fixed 40 ticks and needs a crowd to happen at all, so the drives say how long in terms of the lifecycle rather than in bare counts.
- **Type check:** `bun run lint` green. (Also fixes a pre-existing type error in the walking-record test fixture that was failing it before any of this.)
- **Real browser:** the lab driven in Chromium against the dev server — the phase readout walks `auditioning -> reign -> release` on the first scene, the churny crowd and the single cursor both stay `idle` for their whole run, the ring follows the soloist, and moving a slider mid-scene lands on the promotion in flight. No console errors.
- **Click scanner:** 26 of 30 scenes at 0, including the two new promotion-churn scenes that render whole promotions at the default tuning. See below.

## Not green yet

The full scanner reads **5 clicks**, all on the three mid-scene action scenes:

| scene | clicks |
| --- | --- |
| `sweeps-choral-toggle` | 1 |
| `sweeps-noop-actions` (control, Spencer arrangement) | 0 |
| `sweeps-reset-presence` | 2 |
| `sweeps-noop-reset-shape` (control, default arrangement) | 2 |

The control on the default arrangement reading 2 says those are the harness, not the engine: a `connect` of an already-live node ahead of the render head is the one un-timed effect not yet redirected. The control on the Spencer arrangement reads 0, so the choral toggle's 1 looks genuine.

Driving every tick in lockstep with the render would remove the whole artifact class — the engine's clock would always be the render head — but it surfaces roughly seven further discontinuities that the pre-driven scanner structurally cannot see, since every un-timed call in a pre-driven scene lands at time zero. Those look like real engine faults on paths nothing has exercised before, and chasing them is its own piece of work.

Holding the merge into `sound-notes-prototype` on this.

https://claude.ai/code/session_01BZmBmAFHTJ82EJniMyiChz

---
_Generated by [Claude Code](https://claude.ai/code/session_01BZmBmAFHTJ82EJniMyiChz)_


spencerc99:
Reconciled with the base again (`985688ca`). Two things worth recording, one good and one awkward.

**The harness is now clean.** `53142546` — driving only as far as the suspension's granted head rather than the action's nominal time — closes the last artifact I had left open. Both no-op controls now read **0**:

| scene | before | now |
| --- | --- | --- |
| `sweeps-noop-actions` (Spencer arrangement) | 0 | 0 |
| `sweeps-noop-reset-shape` (default arrangement) | 2 | **0** |

That matters beyond the two clicks: with both controls at zero, anything the action scenes report is now attributable to the engine rather than to the harness. The scenes are trustworthy for the first time.

**The awkward part: the engine fixes on this branch are not exercised by these scenes.** I reverted each in turn and re-scanned. Not one moved the count:

| reverted fix | `sweeps-choral-toggle` | `sweeps-reset-presence` |
| --- | --- | --- |
| *(none — current state)* | 1 | 3 |
| choral `fadeFormants` → `detachFormants` | 1 | — |
| `clearFlourish` disconnecting notes again | — | 3 |
| reset no longer releasing the halo | — | 3 |

Same counts, same timestamps, essentially the same ratios (0.99 vs 1.03 on the toggle). So these scenes were built to prove those fixes and they do not: whatever they are catching is a different fault, and the fixes remain unproven by measurement. They still look right on inspection — disconnecting a bank that is still carrying signal is wrong however little it happens to cost here — but nothing in the matrix currently demonstrates that.

**What is left.** 4 clicks, all at an action's head:

- `sweeps-choral-toggle`: 1, at 8.011s (the first toggle-off)
- `sweeps-reset-presence`: 3, at 8.011s and 16.010s (both resets)

Both scenes click at 8.011s on different arrangements doing different things, and both controls are silent at that moment — so the common factor is something an action itself does at the head, not the particular action. That is the thread to pull next, and it is a clean one now that the harness is out of the way.

I have not touched the engine; the tree is identical to the base.

---

spencerc99 (commented):
## Code review

PR #483, opened 2026-09-09
Preserve participant colors in large archive responses

spencerc99:
Large archive requests silently lost participant colors: a production request for 20,000 keyboard events returned no colors, while all 1,000 matching events in a smaller response had colors. The participant lookup put hundreds of public-key IDs in one URL and ignored query errors.

Load archive colors in batches of 50 IDs. Archive lookup failures now return an error instead of a successful uncolored response. Live broadcasting, event storage, and playback are unchanged.

Verification:

- All 152 Worker tests and the Worker TypeScript check pass. The sandboxed D1 test startup timed out; the complete suite passed with local runtime permissions.
- Real database verification of the changed handler returned colors on all 20,000 keyboard events across 240 participants.
- Automated Chrome verification loaded the production installation typing page with its archive request forwarded to the changed local HTTP handler using the real database. All 22 inspected rendered boxes carried participant colors and randomizeColors was false. The live WebSocket remained connected to production. This verifies the handler and renderer, not a deployed Worker version.
- Production live sample: all 746 events, including three keyboard events, carried colors.

No schema changes or extension release required. A Worker deployment and reload of already-open installation pages are needed for the production fix.

### DIFF
--- a/extension/worker/src/__tests__/participantColors.test.ts
+++ b/extension/worker/src/__tests__/participantColors.test.ts
@@ -1,0 +1,41 @@
+// ABOUTME: Tests bounded participant color loading across large event selections.
+// ABOUTME: Verifies complete coverage, duplicate IDs, and explicit lookup failures.
+
+import { describe, expect, it } from 'vitest';
+import { loadParticipantColors } from '../lib/participantColors';
+
+describe('participant colors', () => {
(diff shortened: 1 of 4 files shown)


spencerc99 (commented):
## Code review

PR #487, opened 2026-09-11
sound: phrase each voice, and a density harness to tune it with

spencerc99:
## Summary

Phases 0 and 1 of the sound upgrade plan, built on `sound-notes-prototype`. Stopping here so you can listen before Phase 2.

**Phase 1 — per-voice phrasing.** Splits every trail voice into instant parameters that keep following motion each frame (gain, brightness, vowel, pan) and one phrased parameter that only changes on an event: pitch.

- Motion is smoothed — speed as an EMA of px/frame, heading as an EMA of the velocity *vector*, weighted by how far each frame actually travelled so a sub-pixel twitch cannot steal the direction of a gesture.
- A note is taken when the heading swings past 30°, at least 200ms after the last one, while the trail is actually moving. Pitch still comes from direction → palette slot → register fold; what changed is that it is chosen once and then held. The per-frame pitch re-read is gone.
- The glide into a note scales with how sharp the turn was: a hard corner near-instant, a gentle curve over ~150ms.
- Every note swells to full over 30ms and then relaxes toward a sustain level over a couple of seconds, on an **articulation node of its own** so the long settle and the control-rate gain never contend for one `AudioParam`. The node sits at unity until a note strikes it, and each attack climbs from the level actually sounding, so re-articulating a settled line is continuous. `ParameterAutomation` gained a `target()` (`setTargetAtTime`) that records its own curve, so holds stay continuous across it.
- The velocity → octave jump is gone. Speed instead crossfades in a quiet ×2 **bloom** layer that stays inside the voice's register, so speed can never introduce an out-of-key pitch, and drives the reverb send inversely (0.4 slow → 0.15 fast).
- **Voice states.** `moving` as above; `lingering` when every point of the last 2s sits inside a small radius but the trail is still moving — the note is held and allowed to develop (vowel closes to a hum, vibrato widens, the octave layer eases in); `resting` after 1.5s genuinely still — fades out over 2s, and resuming motion starts a fresh note rather than continuing the old line.
- The drawn trail's opacity and stroke ride the same articulation value, so the line breathes with the sound. Quantized to 0.02 steps so the trail renderer's own paint cache still does its job.

All of it sits behind a `phrasing` setting — default **on** for the pages and the saved arrangement, **off** for a bare engine — so the previous behaviour stays reachable to A/B against. Turning it off mid-scene hands every sounding voice back to unity and drops its octave layer and velocity reverb, rather than leaving it parked where its last note settled. Every number lives in one tunable file, `shared/sound/tuning.ts`, ready for phases 2-5.

**Phase 0 — density harness.** The tool the rest of the plan gets tuned with, on the sounds playground under the existing replay.

- Recorded trails are `{ id, color, points: { x, y, t }[] }`. Real browsing is split into strokes at the gaps between gestures, each re-based onto its own clock so it can be placed anywhere. Recordings export and import as JSON; a hand-edited file is read back leniently (points sorted and re-based, unreplayable entries dropped).
- Loops them at **1 / 5 / 10 / 30 / 100** at once. Each slot starts at a random offset into its trail and takes a fresh one when its own runs out, so a crowd is never a hundred cursors moving in step.
- Debug readout: energy, chord, progression, full vs pooled voices, and one line per voice — state, smoothed speed, articulation, bloom, pitch, soloist flag. (`pooled` is always 0 until the voice manager lands in Phase 2; the field is there so the readout does not change shape then.)

### Notes and open questions

- **`traceability` interacts with phrasing.** The dial still loosens home-tone bias and swell timing as before; I left it alone rather than folding it into the new note selection. Worth listening to at both ends.
- **`MIN_NOTE_INTERVAL_MS` (80ms) is now dead on the phrased path** — the 200ms turn interval supersedes it. Left in place because the unphrased path still uses it.
- Phase 1e's "remove the ×2/×4 pitch jumps" was already effectively done: `velocityToOctaveMultiplier` was defined but never called. It is still uncalled; say the word and I will delete it.
- Presence (the soloist halo and its dry send) takes priority over the bloom and the velocity reverb for the duration of a reign — two octave doubles on one line read as a chorus rather than as depth.

## Verification

- `bunx tsc -p extension/website` — clean.
- Full extension suite: **181 files, 1387 tests passing**, including 22 phrasing tests, 10 recorded-trail tests, and 8 engine-level phrasing tests. The existing 255 sound tests pass unchanged with phrasing off, which is the regression guard on the old path.
- Real Chromium (`/opt/pw-browsers/chromium`, Playwright) against the dev server at `/sounds/`, driving the harness end to end: audio context `running`, energy scaling **0.013 → 0.244 → 0.597** across 1 / 30 / 100 trails, voices reporting `moving` and `lingering`, and bloom tracking speed. Toggling `phrasing` off mid-scene returns every voice to articulation 1.00 and bloom 0.00 with the scene still running and no page errors.

Note: `PR Validation` only triggers on PRs into `main`, so CI here runs the publish/preview checks only. The local run above is the actual coverage for this branch.

Harness at 30 trails — phrasing on (lines breathe with their voices):

Density harness at 30 trails with phrasing on

The same scene with phrasing toggled off mid-run (uniform lines, full level):

Density harness at 30 trails with phrasing off

The per-voice readout:

Per-voice readout: energy, chord, progression, and one line per voice

**What I could not capture:** the articulation → drawing tie-in on its real surface, the archive/portrait pages, because the wewere.online worker is unreachable from this sandbox (proxy returns 403 on `api.wewere.online`) and those pages have no offline data path. The harness canvas above is driven by the same `getArticulation` readback and the same `articulationBreath` mapping, and there is a test asserting `AnimatedTrails` asks the engine per visible trail as it draws — but you should give the archive a look before this merges.

Not an extension feature — this is `extension/website/**` (the sounds playground and the shared visualization code), which deploys on its own and gets no `PENDING.md` bullet or extension release. No `packages/` change, so no changeset, no docs audit, and no starter-template update.

https://claude.ai/code/session_01Nope3LfBbooHZPQbXaN4LR


spencerc99 (commented):
## Code review

### Medium

- [`extension/website/shared/sound/SoundEngine.ts:3393`](https://github.com/spencerc99/playhtml/blob/614746e8175bb88e8f372f67665871ef9cfa17d1/extension/website/shared/sound/SoundEngine.ts#L3393) — The new articulation node starts every voice at `articulationRestLevel` (0.6) even when `phrasing` is disabled. Because a bare engine defaults `phrasing` to false and the off state never calls `articulate`, the supposedly unchanged fallback path is now permanently 40% quieter; toggling phrasing off also leaves an active voice at whatever articulation level it last reached. Initialize/bypass this node at gain 1 while phrasing is off, and explicitly restore active voices to 1 when the setting is disabled.

- [`extension/website/shared/sound/phrasing.ts:94-102`](https://github.com/spencerc99/playhtml/blob/614746e8175bb88e8f372f67665871ef9cfa17d1/extension/website/shared/sound/phrasing.ts#L94-L102) — Normalizing every nonzero delta before the heading EMA gives a sub-pixel movement exactly the same directional weight as a real moving frame. After a fast run, the speed EMA remains above `moveMinSpeed` for several frames, so a few tiny off-axis or opposite jitter samples can rotate the held heading past 30° and trigger a false note even though the cursor is effectively stopped. Smooth the scaled velocity components directly, or gate/down-weight heading updates by movement magnitude, and cover small off-axis jitter after a fast run.

spencerc99:
Both findings were real. Fixed in `8a9e1d1c`.

**Articulation node parked below unity.** Confirmed — the node was created at the settled level, and with phrasing off nothing ever articulates it, so the whole fallback bed sat 40% quieter for the life of the scene. Three changes:

- The node is created at unity. Only a note strikes it, and a note only happens with phrasing on.
- The envelope is now modelled from the level actually sounding at the strike (`articulationFrom`) rather than from a fixed floor, so it matches what the audio ramp does — which also makes re-articulating a settled line continuous instead of a dip followed by a jump. `articulationRestLevel` is gone; the pre-note value is simply 1.
- `setConfig({ phrasing: false })` now hands every sounding voice back to unity, drops its octave layer and velocity reverb, and clears its phrasing state — immediately rather than on the next tick, which a paused canvas may never run.

My old "leaves the unphrased path exactly as it was" test only asserted that no envelope curve ran, which the bug passed cleanly. It now also asserts that no gain node sits parked under unity, and I confirmed it fails against the old initializer before keeping the fix. A second test covers the mid-scene toggle.

**Heading EMA over unit vectors.** Also confirmed, and worse than it reads. Driving a run at 12 px/frame and then feeding 0.05 px/frame of off-axis jitter, the held heading rotated **76°** and fired a false note from frame 12 onward — exactly the drain window you describe, where the speed EMA is still above `moveMinSpeed` but the cursor has stopped.

Fixed by smoothing the scaled velocity components directly instead of normalizing them, so each frame is weighted by how far it actually travelled. Same jitter sequence now moves the heading **5.8°** and takes no note. `headingAngle`'s floor moved with it — the vector now carries magnitude, so the floor is in speed units (a tenth of `moveMinSpeed`) rather than a unit-vector epsilon.

Verified: typecheck clean, 1387 tests passing, and re-driven in Chromium at 30 trails — with phrasing on, articulation varies 0.78–0.99 and bloom tracks speed; toggled off mid-scene, every voice reports articulation 1.00 and bloom 0.00 with no page errors. Screenshots in the PR body are updated to this commit.

---

spencerc99 (commented):
## Code review

PR #488, opened 2026-09-13
extension: play live scrolling and typing while they are still happening

spencerc99:
## Summary

The installation's scrolling and typing screens withheld a recording until its last event was 35 seconds old (`collectInstallationRecordings`). Because a scroll group only breaks after 15 minutes of inactivity on a page (`SCROLL_SESSION_THRESHOLD`) and typing merges within 35 seconds (`TYPING_EVENT_MERGE_THRESHOLD_MS`), an active session kept resetting that clock — so someone browsing in front of the screens saw nothing at all until they stopped and sat still for 35 seconds. That was the "even after a minute, nothing shows up."

This admits a recording the moment it has any footage, and keeps extending it on screen as more arrives:

- **Reservoir** (`useInstallationRecordings`): no settle gate. A recording is retired only once no event in it came from the archive *and* it has been quiet for `LIVE_RECORDING_RETENTION_MS` (60s), so live-only footage lingers while it still reads as "now" and then gives way to the archive as its source. Retirement is clock-driven rather than arrival-driven, so the result reports `hasRetirableRecordings` and the reservoir keeps re-evaluating (every 1s) until nothing is waiting. The signature is now one entry per recording rather than per event, so ticking more often does not get more expensive.
- **Extend in place** (`AnimatedScrollViewports`, `AnimatedTyping`): a window already on screen is handed longer footage rather than ending at whatever it held when admitted. This works because an animation's `durationMs` and its timeline's `timeRange` grow by the same amount, so `getViewportFrame`'s playhead carries on instead of jumping — the same window keeps scrolling, and the same box keeps typing. A recording keeps a stable id throughout (the id of its first event), which is what makes the match possible.
- **Retaining footage** (`pruneRetainedLive`): live events are freed a whole recording at a time — what no surviving recording still claims, then the oldest recordings entire if that leaves too much. Never by age: an unbroken run has no age bound (500ms scroll samples against a 15-minute silence threshold), and losing a recording's first event would change its id and restart the window on screen.
- **Playback queue**: recordings carry a `version`, so grown footage is offered again as live material rather than being treated as already played, and `current(id)` hands players the newest version of what they are already showing.

Latency for scrolling and typing goes from *unbounded while you are still active* to about a second.

### Not included

Cursors are unaffected — they already bypass this path entirely via `continuousLiveTrails` on the `cursors`/`follower-*` profiles. Two related gaps I found while tracing this but did not touch here, happy to do either: `clicks` is not on that continuous path, and `liveChapterIsReady` requires ≥2 distinct participants for trails, so a single visitor is invisible to every non-profile screen.

### Repo checks

No changeset — nothing under `packages/` changed. No `apps/docs/` audit — this is the extension website, not the playhtml library. No `PENDING.md` bullet — per `extension/CLAUDE.md`, changes under `extension/website/**` deploy on their own and do not get extension release notes. No manifest permission changes.

## Verification

- `bun run test:extension` — 171 files, 1069 tests passing (was 1053; 16 added/rewritten).
- `bun run lint` — clean across all packages.
- Rewrote `installationRecordings.test.ts` for the new semantics: immediate admission, in-place growth, retirement at the retention boundary, archive-backed footage never retiring, and the clock-driven re-evaluation guard.
- Added retention coverage that runs the reservoir's own loop (retain → collect → prune, feeding each pass's pruning into the next) across a 20-minute unbroken run and asserts the recording stays whole. Verified red against age-based eviction, so it pins the bug rather than merely passing.
- Added queue coverage for growth (a grown recording jumps the archive again; `current()` returns the newest version) and three `AnimatedScrollViewports` tests for the extension invariant — that a recording handed more footage does not move the position already on screen, and carries on past where it would have ended.

**Real-browser verification** (not an extension feature, but this is the behavior the whole change is about, so I automated it):

Temporary Playwright harness driving the `extension/website` dev server against a stub Worker serving an empty archive plus a `/stream` WebSocket I emitted frames into by hand. Chromium 1194 headless, 1440×900. Re-run after the retention fix with identical results.

`/installation/live/?screen=scrolling`:

| | result |
| --- | --- |
| partial footage admitted | **0.9s** after the first frame (previously: never, while still scrolling) |
| distinct recordings while growing | **1** — same `data-scroll-recording`, scroll transform advanced `translate(0 -22.5)` → `translate(0 -276.5)` across six later frames |
| at 30s quiet | still on screen |
| at 75s quiet | retired |

`/installation/live/?screen=typing`: admitted in **0.9s** mid-word (`"the quick br"`), then the *same* box continued into the appended text (`"the quick brown fox jumps over the l"`, still in `phase: typing`) with a live caret, rather than restarting as a second box; retired after the quiet window.

The retirement half initially did not fire — the first browser run caught that my tick guard skipped exactly the pass that should have retired the recording. Fixed via `hasRetirableRecordings` and covered by a unit test.

**Seeing it yourself:** screenshots of each stage were captured during verification but cannot be attached to GitHub from the session this was built in. The Cloudflare branch preview runs this build, so the live result is available directly — [scrolling](https://claude-extension-delay-live.we-were-online-website.pages.dev/installation/live/?screen=scrolling), [typing](https://claude-extension-delay-live.we-were-online-website.pages.dev/installation/live/?screen=typing). With installation mode on, scroll or type and the footage should land in about a second and keep going.

## Notes for review

- **Compression vs. real time.** Live recordings still play at `SCROLL_TIME_COMPRESSION` (10×), so a growing window catches up to the incoming data and briefly parks before more arrives. It reads as live and is honest about the footage it has, but playing live-growing recordings at real time is a plausible follow-up if the pacing feels stuttery on the wall.
- **`scrollRange` rescales as a recording grows**, so a window can shift slightly when the person scrolls past a new extreme. Inherent to showing partial footage; smoothable later if it reads badly.

https://claude.ai/code/session_01TuCMGKxmQqArFafv4B61Bq


spencerc99 (commented):
## Code review

### Medium

- `extension/website/shared/hooks/useInstallationRecordings.ts:21` — The 10-minute age cap can truncate a still-active scroll recording before the 15-minute `SCROLL_SESSION_THRESHOLD` can close its group. If archive ingestion is delayed or unavailable and scrolling continues past 10 minutes, the first live event is deleted, so `useViewportScroll` derives a new `eventId`; the active viewport can no longer be extended and the remainder may be queued as a second/restarted window. Preserve the first event and evict only completed groups, or make the retention bound exceed the grouping threshold plus archive lag.
- `extension/website/shared/components/AnimatedScrollViewports.tsx:311` — This changes visible installation behavior, but the PR provides no attached screenshots despite the repository requirement and the body saying screenshots were captured. Attach the scrolling and typing stage screenshots used for the browser verification so reviewers can validate the user-facing result.

spencerc99:
Both findings addressed.

**1. Age-based eviction truncating a still-active recording — fixed in 19bc24c.**

Confirmed, and worse than a tuning problem: no fixed age bound can work here. A recording holds one unbroken run, scroll samples land every 500ms, and `SCROLL_SESSION_THRESHOLD` only breaks a group after 15 minutes of *silence* — so an unbroken run can span arbitrarily long. Raising the bound past 15 minutes would just move the cliff. The comment I had on `RETAINED_LIVE_MAX_AGE_MS` claimed precisely the guarantee the code failed to provide.

Eviction is now per-recording (`pruneRetainedLive`): drop what no surviving recording still claims, then, only if that still leaves too much, drop the oldest recordings entire. Age no longer comes into it, so the first event survives for as long as the person keeps going and the id stays stable.

Covered by a test that runs the reservoir's own loop (retain → collect → prune, feeding each pass's pruning into the next) across a 20-minute unbroken run and asserts the recording stays whole with `ev_0` still first. I checked it goes red against the old age-based eviction, so it pins the reported bug rather than just passing. Two more cover the retired-recording case and oldest-recording-first eviction at the cap.

**2. Screenshots.** Fair, and the PR body was misleading — it said screenshots were captured (true) in a way that read as though they were attached (they were not). I've reworded it.

I can't attach images to GitHub from this environment; the screenshots went to Spencer directly in the session. What I can offer reviewers instead is the live result rather than a picture of it — the Cloudflare branch preview runs this build:

- https://claude-extension-delay-live.we-were-online-website.pages.dev/installation/live/?screen=scrolling
- https://claude-extension-delay-live.we-were-online-website.pages.dev/installation/live/?screen=typing

With installation mode on, scroll or type and the footage should land in about a second and keep going, rather than waiting for you to sit still.

Re-verified after the fix: full suite 1069 passing, `bun run lint` clean, and both browser flows re-run unchanged — scrolling admitted in 0.9s as one recording (`translate(0 -22.5)` → `translate(0 -276.5)`), typing admitted mid-word and continuing into the appended sentence in the same box, both retired after the quiet window.

---

spencerc99:
## Code review

### Medium

- `extension/website/shared/components/AnimatedScrollViewports.tsx:311` — The previous screenshot finding remains unresolved. This callback implements the user-visible in-place scrolling extension, and the PR also changes the typing installation surface, but the PR body and author response explicitly confirm that no screenshots are attached. The repository requirement says every user-facing change must include screenshots of the finished real surface and key interactive states in the handoff and PR; live preview links do not replace that evidence. Attach the captured scrolling and typing stage screenshots to the PR so reviewers can validate the finished states.

spencerc99:
On the screenshot finding — stating the constraint plainly rather than repeating the earlier answer, since this is the second time it has been raised.

The screenshots exist. All seven stages of both flows were captured during verification and delivered to Spencer in the session that produced this branch. What I cannot do is attach them here: GitHub image attachment goes through the web UI's upload endpoint, which the REST API does not expose, and this session has no other route to it. That is an environment limit, not an omission or a judgement that the requirement does not apply.

Spencer has the files and has decided to leave it here rather than have me commit verification artifacts into the source tree (the repo's only media convention today is changelog media, and the raw captures are ~2.5 MB each). So the requirement stands formally unmet on this PR, deliberately and with the author's knowledge, rather than quietly.

What a reviewer can use in the meantime, for whatever it is worth alongside that gap:

- the branch preview, which runs this exact build — [scrolling](https://claude-extension-delay-live.we-were-online-website.pages.dev/installation/live/?screen=scrolling), [typing](https://claude-extension-delay-live.we-were-online-website.pages.dev/installation/live/?screen=typing)
- the measured DOM state from the automated run, in the PR body: one `data-scroll-recording` throughout growth with its transform advancing `translate(0 -22.5)` → `translate(0 -276.5)`, and one `data-typing-recording` continuing from `"the quick br"` into `"the quick brown fox jumps over the l"` while still in `phase: typing`

I am not going to re-litigate this on further sweeps; the position above is the standing answer.

---

PR #496, opened 2026-09-17
playhtml: reduce repeated subscriber work

spencerc99:
## Summary

- Keep primitive page-data channels attached to their stable parent observer so synchronous writes produce one final notification.
- Suppress `users.onChange` notifications when cursor positions change but the user roster and identities do not.
- Document page-data notification batching and add a patch changeset.

## Why

The weekly library performance audit found two reachable sources of repeated subscriber work. Primitive page-data writes detached and reattached their observer for every replacement, then notified every listener synchronously. Cursor position updates also notified every `users.onChange` subscriber even though the returned `User[]` had not changed. The latter caused every `usePresence` hook to reread the local identity on every cursor tick.

## Measurements

Measured in Chromium 147.0.7727.15 with seven fresh-room runs per scenario against built package artifacts.

- 1,000 primitive page-data writes with one listener now produce 1 callback instead of 1,000. With 100 listeners, they produce 100 callbacks instead of 100,000.
- 10,000 primitive writes fell from 246.0–250.7 ms to 215.9–217.6 ms.
- At 1,000 `usePresence` hooks, 20 frame-paced cursor ticks now perform 0 identity reads instead of 20,000.
- Active cursor-update CPU fell from 7.9–9.9 ms to 1.9–2.8 ms across those 20 ticks, matching the 1.5–2.7 ms baseline range.
- The public PlayHTML ESM path grows by 269 raw bytes, 108 gzip bytes, and 126 Brotli bytes.

The runtime baseline was `262d360e37da2b6c8f48a89af6592f6f91afecd1`. Current `main` adds only extension files relative to that SHA, so the audited package artifacts and paths are unchanged.

## Verification

- `bun run build-packages`
- `bun run test:common` — 18 tests
- `bun run test:playhtml` — 540 tests
- `bun run test:react` — 56 unit tests, 3 integration tests, and type tests
- `bun run check:packages`
- `bun run test:docs` — 52 tests
- `bun run check:docs` — static build passed
- Seven-run Chromium page-data and React matrices
- Seven-run frame-paced React CPU comparison

## Release notes

- Adds a `playhtml` patch changeset.
- Updates the page-data guide and API reference because notification batching is observable behavior.
- Starter templates are unchanged because no API signature or usage changed.
- Screenshots are not applicable because this change has no visual surface.


spencerc99 (commented):
## Code review

### Medium

- `packages/playhtml/src/page-data.ts:212-216` — Replacing a primitive (including `null`) root with an object leaves this channel’s observer attached shallowly to the parent page-data map. Later writes take the object path through `deepReplaceIntoProxy` at lines 205-208, which mutates the child without changing the parent key, so `onUpdate` stops firing. This is a supported path—`PageDataChannel<Item | null>` is explicitly type-tested in `packages/common/src/page-data.types.ts`. Rebind the observer when the root changes representation, or observe the parent deeply while preserving the new microtask coalescing.

spencerc99:
## Code review

PR #499, opened 2026-09-19
Add durable Internet Commute curation and policy checks

spencerc99:
## Summary

Recover the Internet Commute curation desk and durable catalog on current main. Import audit evidence, review public editorial candidates, and save page/hostname/site decisions separately from machine suggestions.

This is the foundation for a later Jev classification experiment. The existing optional suggestion provider is Workers AI/Llama, not Jev. Selecting an undecided candidate automatically requests a suggestion when that binding is configured; saving a human policy remains explicit.

## Review fixes

- Preserve current main's admin authentication, train boarding, and activity counts while resolving the branch conflicts.
- Require authentication for the uncurated review feed.
- Replace policy scope atomically, so an unsuccessful save preserves the existing restriction.
- Reject protected URLs and send only projected evidence fields to suggestions; omit browsing identifiers and observations.
- Bind cached suggestions to an evidence hash and accept the audit sampler's full bounded output.
- Detect quoted login password fields and treat interrupted inspection responses as unknown.
- Apply D1 migrations before deploying the Worker.
- Make the curation desk usable on narrow screens.
- Check fallback train destinations against durable policies; fail closed with 503 if fewer than two policy-checked stops are available.
- Cancel stalled image body readers explicitly at the download deadline. Full-suite validation exposed an unchanged fingerprinting path whose abort signal could fire without settling the reader under garbage collection; regression coverage uses real stalled HTTP responses, including partial bodies.

## Verification

- Full Worker suite: 25 files, 223 tests, including real isolated D1 tests for import idempotence, policy precedence, rollback, and cache behavior.
- Curation/admin/catalog-generator suite: 8 files, 43 tests. Worker and website typechecks; package build; Worker dry-run bundle; full website build.
- Built-page Chromium automation against the actual local Worker and isolated D1: rejected credentials, audit import, no policy mutation on import, saved decision, reload persistence, atomic hostname-to-page replacement, desktop/mobile screenshots, and lock.
- No mocked browser/API responses. The imported artifact contains three public example pages and simulated test attention, not private browsing records.
- Browser HTTP boarding checks against the built Worker, real isolated D1, Durable Object, and rate limiter: hiding both fallback destinations returns 503; removing those test restrictions returns 200 with the allowed destinations. The missing live feed is intentional in this isolated environment.

The local verification environment has no production credentials or AI binding. Live production history and model inference are not claimed as tested. Screenshots are supplied in PR Evidence. The full website build reports its existing large-chunk warning.

## Deployment and data

Migrations add the place catalog, placement levels, suggestion cache, and evidence hash. The deploy workflow applies pending D1 migrations first. Existing original-worktree D1 state was not modified. No production migrations, deployment, or Jev calls were performed during preparation.

Human routing choices remain distinct from evidence and suggestions. Deterministic URL restrictions still apply before promotion. A public-page inspection or model confidence is not proof of privacy or safety.

Policy checks apply when assembling train routes. Existing train assignments retain their already-created routes; this does not retroactively reroute boarded trains.

No published package API changed, so no changeset, public library docs, or starter-template changes are needed. No extension manifest permissions changed. These website/Worker changes do not require an extension-store release.


spencerc99:
## Preview evidence

**Authenticated curation desk: public evidence imports do not change policies; explicit routing decisions persist across reload; mobile layout is verified.**

[Open preview](http://127.0.0.1:5208/commute-curation/) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-mu7qy55q-680e0b54) · Commit `dfcf069`

### Locked curation desk

Locked curation desk

Authentication is required to read evidence or change policies.

1440×1000

### Evidence imported without policies

Evidence imported without policies

Three public sample pages were imported; zero human policies were created.

1440×1000

### Explicit decision survives reload

Explicit decision survives reload

The page-scoped policy persists after atomic replacement of the hostname policy.

1440×1000

### Mobile curation flow

Mobile curation flow

The desk fits a 390-pixel viewport with no horizontal overflow.

390×844

spencerc99 (commented):
## Code review

- **Medium — The curation desk defaults to the filtered public feed** (`extension/website/commute-curation/main.tsx:363`): `COMMUTE_REVIEW_URL` is optional and has no repository default, so normal builds fall back to `/commute/recent` and omit the admin token. That public endpoint applies existing policies and truncates the candidate set, which means hidden/scenery decisions can disappear from the desk after reload and reviewers cannot audit the full uncurated queue. Derive the review endpoint from `WORKER_URL` by default (while retaining an override if needed) and always send the bearer token when loading this desk.

PR #502, opened 2026-09-20
playhtml: allow clearing nullable page data

spencerc99:
A page-data channel holding an object or array could not be cleared to `null`: direct writes were sent through an object mutator, and callback returns were discarded. Nullable shared selections now clear with either `setData(null)` or `setData(() => null)`, notify subscribers, and accept another selection afterward.

The setter replaces the stored root when an in-place update cannot represent the next value. Object draft mutations retain their existing behavior, including terse callbacks such as `draft.count++`.

Validation:
- Core suite: 52 files, 540 tests passed, including nullable object/array resets, subscription updates, and synchronization into a second real Yjs document.
- All published packages built successfully.
- Chromium with the built package and a loopback Yjs server: two connected pages selected, cleared directly, reselected, cleared through an updater, edited after a peer clear, and reloaded successfully; no page errors. Screenshots: [selected and cleared states](https://pr-evidence.spencerc99.workers.dev/review/ev-mua3p1vu-e3ed1327).
- Docs built successfully. Build warnings remain for the existing fire-hydrant asset reference, circular React chunks, and large chunks.
- Node 26 tests used `NODE_OPTIONS=--no-experimental-webstorage` to allow jsdom to supply browser storage.

Includes a patch changeset and page-data documentation. Starter APIs, dependencies, exports, and persisted data fields are unchanged, so neither starter nor a tarball dependency simulation needs updates. This intersects PR #496's page-data setter/observer file; #496 does not address nullable root clearing.

This PR fixes audit finding 4. The React default-data API decision in finding 5 remains separate.


spencerc99 (commented):
## Code review

PR #503, opened 2026-09-20
extension: keep recording typing after idle pauses

spencerc99:
Typing after a five-second pause in the same field stopped being recorded. The collector cleared its field metadata after the first burst, so later bursts could not flush and could be attributed to the next field instead.

Keep the focused field's metadata between bursts and clear it once no field remains focused. Add regressions for repeated idle flushes, final flush on disable, and switching fields. Include a public release note because this affects ordinary typing sessions.

### Verification

- Both regression tests failed before the fix and pass afterward; all six KeyboardCollector tests pass.
- Package builds, production Chrome extension build, and extension TypeScript check pass.
- Real production extension in isolated Chromium 147.0.7727.15: complete onboarding through the UI with Local keyboard collection and cadence-only redaction; focus/type/idle/type/idle without refocusing; switch fields; type into a password field; reload and resume typing. Background API reads verify five persisted keyboard events, correct field association, masked text, and zero password events. No mocked collector, browser storage, or extension messaging.
- Full extension suite: 1,100 tests pass; one unrelated image download-timeout test fails at 15 seconds. The same failure reproduces from a clean main archive at `011240f2` with a fresh locked dependency install and package build. The suite is not fully green.

No published package API, starter template, manifest permission, or public documentation changes are needed. This PR does not change typing privacy rules.

### Browser evidence

[Verified browser evidence for this commit](https://pr-evidence.spencerc99.workers.dev/review/ev-mua3ovi2-35dd9584).

Second typing burst without refocusing:

Second typing burst in the same field

Persisted, masked typing in the real extension overlay after reload:

Persisted typing after reload


spencerc99 (commented):
## Code review

PR #504, opened 2026-09-20
Fix moderation removal for dotted data keys

spencerc99:
## Summary

Moderation can remove records beneath element IDs and data keys containing periods. For example, `chat.v1` and a nested `chat` → `v1` object now produce distinct selection keys, even when both contain identical messages. Deletion preserves the neighboring record, the content-hash stale guard, duplicate-target handling, and descending index removal.

The targeting key escapes its separators; the display path and ordinary keys retain their existing shape. This is an internal administrative path, with no published package API or starter-template changes and no changeset required.

## Verification

- Added regressions first: dotted/nested path collision and special/empty path segments failed on the base implementation.
- 37 tests pass across moderation, real Y.Doc persistence round trips, admin snapshot creation, and document comparison.
- `bun run check:partykit` passes.
- Automated the real local `/admin.html` flow in Chromium against the actual PartyKit Worker, isolated PostgreSQL 15, and PostgREST. Selected exactly one of two identical messages under literal dotted versus nested keys, removed it through the real authenticated admin route, then reloaded and read the persisted database snapshot. The nested and unrelated records remained intact.

All browser data and credentials were synthetic and local. No production requests or data changes. Browser evidence uses the local Vite admin surface and real Worker/database; deployed behavior has not been verified.

## Screenshots

Captured from the verified local admin flow at this commit.

Dotted record selected independently

Persisted deletion preserves adjacent records


spencerc99 (commented):
## Code review

PR #505, opened 2026-09-20
Fix room switching after client-side navigation

spencerc99:
## Summary

Switch rooms when the browser commits a navigation. Previously, `pushState()` from `/first` to `/second` triggered room lookup while the URL still pointed at `/first`, leaving shared edits attached to the previous room. The listener now uses `currententrychange`; `popstate` remains available for browsers without the Navigation API.

Adds a real Chromium regression to PR validation covering pushState, replaceState, back, forward, cancelled navigation, and listener cleanup. The regression failed on the original first pushState before the fix.

Includes a `playhtml` patch changeset and updates the navigation guide. No public API changes: the HTML and React starters already use the supported initialization APIs and need no edits.

## Verification

- `bun run smoke:navigation` passed.
- Built-package end-to-end verification passed with the actual local PartyKit Worker and isolated PostgreSQL/PostgREST. Four independent browser contexts confirmed destination page-data and element writes after pushState/replaceState, unchanged state in other rooms, and restored room state on back/forward. No browser errors or warnings.
- `bun run build-packages`, `bun run check:playhtml`, and `bun run build-site` passed.
- Core: 537 tests. React: 56 unit tests, 3 integration tests, and type tests. Docs: 52 tests.
- Docs tests retain existing Sass legacy API and Lit dev-mode warnings; the site build retains existing Sass deprecation, circular React chunk, and chunk-size warnings.

Runtime verification uses a local fixture importing the built package and real local services. Production deployment and long-term persistence were not tested. Router integrations that swap DOM asynchronously should continue to call `handleNavigation()` after their swap as documented.

## Visual evidence

Captured against commit `3847f9efb23bc604c38fd857c1a7477274f1dfaf` using the built package and local services. The local preview URL is not a hosted deployment. [Evidence dashboard](https://pr-evidence.spencerc99.workers.dev/review/ev-mua3rbw8-4036eff7).

Destination room after pushState, with its own saved route and toggled element:

Destination room after pushState

Back restores the first room's separate state:

Original room restored with Back


spencerc99 (commented):
## Code review

PR #506, opened 2026-09-20
react: require values for element defaults

spencerc99:
React children need initial data before their DOM exists. A DOM-dependent `defaultData` callback left `data` undefined on the first render, so a child reading `data.count` crashed before initialization could run.

Require values for React `defaultData` and `myDefaultAwareness`, remove deferred DOM initialization, and reject callbacks before rendering children. Type tests cover explicit and inferred callback props. Built-in React capabilities validate their shared initializer defaults at the React boundary.

Breaking change: derive defaults from props, either directly (`defaultData={{ count: initialCount }}`) or through `withSharedState(props => ({ defaultData: ... }), render)`. A major `@playhtml/react` changeset records this contract change. Vanilla DOM-dependent initializers and saved data shapes are unchanged.

Migrated the docs scroll rail awareness into its outer config factory. Updated the React reference and both starter templates: React derives the reaction count from a prop; vanilla derives it from an existing element attribute.

Validation:
- Public type checks passed, including rejection of DOM callbacks and acceptance of props-based configuration.
- Full React suite: 56 unit tests and 7 integration tests passed. Tests validate expected server-render warnings; no uncaptured test errors.
- All published packages built; generated declarations contain the value-only contract.
- Docs built. Existing warnings remain for an external asset reference, circular React chunks, and chunk sizes.
- Chromium against built packages and a loopback Yjs server: verified first-render data/awareness, increment, reload with saved state, and both starter reaction interactions. The built docs React reference and its scroll rail also hydrated without page errors. No page exceptions; starter external fonts/images reported loading errors.

Dependencies and exports are unchanged, so no dependency-boundary tarball simulation was needed. Browser captures: [first render, update, both starters, and docs scroll rail](https://pr-evidence.spencerc99.workers.dev/review/ev-mua4lfz4-1ecb7ad2). Nullable page-data clearing remains in separate PR #502, which is mergeable with passing checks.


spencerc99 (commented):
## Code review

apps/docs/src/components/react/DocsScrollRail.tsx
```diff
@@ -53,10 +53,10 @@ function computeSectionTicks(): SectionTick[] {
 }
 
 export const DocsScrollRail = withSharedState<{}, RailAwareness>(
-  {
+  () => ({
     defaultData: {},
-    myDefaultAwareness: () => ({ color: readCursorColor(), scroll: 0 }),
-  },
+    myDefaultAwareness: { color: readCursorColor(), scroll: 0 },
+  }),
```
spencerc99: wait why did we need to do this? this seems very ugly

PR #507, opened 2026-09-20
Add source filters and inline search to Internet Scraps

spencerc99:
## Summary

Add Found on and Type popovers to Internet Scraps, with search that expands inline beside them. Each selected source appears as its own clickable chip. The dock fits its contents, and the browsing-mode and layout toggles sit together. Source filters accept domains and URL/path prefixes using the portrait archive's matching rules; search matches saved page titles, URLs, image alt text, and button text.

Group photos before filtering so a match at any saved source retains the full encounter history. Buttons, icons, and cursors are filtered before deduplication, so an earlier matching occurrence remains searchable even when the latest occurrence came from another page. Filters carry across Pile/Grid and Drift/Archive, reset the archive to its first result, and show an empty state when nothing matches. Drift pauses while the controls have focus. Selections reset on page reload.

## Validation

- 75 relevant tests pass, including source-aware search, photo provenance, controls, layouts, scrolling, tide behavior, and the lightbox.
- Extension and website type checks and the production Chrome extension build pass.
- The isolated built-extension browser flow covers real collection and restart, source/path filtering with complete history, multiple domains, empty type results, inline search, all four mode/layout combinations, mobile controls, and repeated non-image scraps across source paths. Screenshots are attached through PR Evidence.

Scraps is beta, so this does not add a public release-note entry. No extension permissions, backend contracts, storage formats, or published-package APIs change; library documentation and starter templates are unaffected.

PR #461, opened 2026-09-06
Sound overhaul prototype: generative orchestra for cursor trails

spencerc99:
## What this is

A full rework of the generative sound in the movement visualizations (`extension/website/shared/sound/` + the `/sounds/` playground), developed over ~5 weeks of listening-driven iteration.

## Guiding principle

**Orchestra legibility**: a watcher should be able to derive what causes each sound, like watching an orchestra. Every sound maps to a visible event; bursts absorb into weight instead of repetition.

## What ships

**Engine** (`SoundEngine.ts` + `scales.ts` + `instruments.ts`)
- Two modes: sustained (default) and spotlight (soloist lifecycle: silent audition, understated entrance, minimum reign, cooldown, population floor; presence voice keeps the trail's own register and steps forward via brightness/closeness/vibrato with an octave halo)
- Chord rotation across 5 selectable diatonic progressions with voice-led transitions; per-trail voices (color-derived SATB registers, home chord tones, detune/vibrato fingerprints); swells; choral formant timbre; energy arc; bass pedal; cantus firmus
- Accents: door-chime arrivals/departures, pizzicato/timpani click+hold voices, crossing tension/merge, data-driven navigation gong (off by default)
- Click-storm admission control: overflow clicks are dropped before allocation, never evicting ringing notes
- Per-voice reverb sends; real-time perf work (changed-target automation guards, lazy oscillators, node pooling, parked modulation)

**Shipped default**: sustained mode, navigation sounds off — the calm arrangement (chords, voices, swells, chimes, bass pedal, cantus on).

**Playground** (`/sounds/`): real-event replay at archive density, unified layer panel with mixer solo/mute, soloist lab with lifecycle sliders + scripted scenarios, audition library, config persistence, dev performance readout.

**Surfaces**: live portrait and installation cursor screen auto-apply the arrangement (no UI); archive gets the dev panel behind `?sounddev=1`. Production pages without the flag are unchanged.

## Verification

- `sound-regression` CI job runs an offline click scanner: renders 33 deterministic scenes (dense, live-cadence, churn, storm, mid-scene toggles) and asserts 0 audio discontinuities via order-8 linear-prediction error
- 1,406 extension tests passing; lint clean
- Every fixed bug class (envelope truncation, ramp-inside-ramp, eviction cuts, budget starvation) has a scanner scene + regression test
- Real-ear verification throughout by Spencer on the playground, portrait, and deployed previews

## Notes

[...]

PR #508, opened 2026-09-21
extension: collect headings, scrap position, and button backdrops

spencerc99:
## Summary

Internet scraps collects three more things while you browse, all local-only like the rest of scraps:

- **Headings** (`h1`-`h3`) as a new `heading` scrap kind: the page's own wording, the level, and a typographic allowlist (font family/size/weight/style, text color, letter-spacing, text-transform, line-height). No HTML and no backgrounds are stored. Skipped: under 2 or over 120 characters, hidden, never actually seen. Deduped per domain by lowercased text, level-insensitive.
- **Where each scrap was seen**: an optional `position` (`pageX`, `pageY` of the element center in document coordinates, plus `pageWidth`/`pageHeight`) on every kind. Cursors use the pointer position.
- **The backdrop a button sat on**: when a button's own background does not cover it, the first opaque ancestor color is saved as a separate `backdropColor` (the button's own styles stay truthful) and painted as a snug patch behind it. Fixes light-text buttons redrawing invisible.

And one thing it stops collecting: **"buttons" with no button look**. A control with no background, gradient, border, shadow, or icon is plain text tagged `role="button"` and is skipped at collection.

## Why

Groundwork for the scrap collage builder: headings are words-material, position enables "what was found here" tools and page-visit portraits, and neither position nor backdrop can be backfilled, so every day uncollected is lost. In a real 5,000-scrap export, 964 of 1,961 buttons had no visible chrome (about 550 of them icon-less prose like "willhess17 4 years ago (edited)"), and 297 light-text buttons rendered as nothing.

## Notes for review

- All new fields are optional on stored records because existing scraps lack them; no defaults are invented for old records and nothing is filtered at read time. `heading` is an additive union variant. No migration needed.
- Position, backdrop, and heading level are not part of dedup identity.
- Pale headings can be faint on the scraps paper. Accepted: headings are material, not a display surface.
- Fixed along the way: `innerText` stored already-transformed text so `text-transform` applied twice (now `textContent`); archive tiles squashed wrapped headings.
- Known, not fixed here: `getButtonText` has the same latent `innerText` issue (harmless today), buttons overflow their tiles in archive pile (pre-existing on main), and `extension/website/scraps-{grid,pile,preview}` are outside the website tsconfig so their `kind` switches are not type-checked.
- No manifest permission changes. No worker or `@playhtml/extension-types` changes (the worker does not validate scrap kinds). No changes under `packages/`, so no changeset, docs, or starter-template updates.
- `extension/PENDING.md`: one bullet for heading collection (SCRAPS is reachable at beta). None for position, backdrop, or the bare-button rule.

## Verification

`bun run test:extension` 1150 passed, `check:extension` clean, `check:extension-website` clean, `build-extension` clean.

**Real-browser automation** (temporary Playwright harness, not committed): the built `chrome-mv3` extension in an isolated headless Chromium on a fresh profile, all network blocked except a local origin. Scraps enabled through the real options-page checkbox, then a 900x4600 test page is visited and scrolled in stages so each element passes the visibility gate: three headings (one with its own opaque background), a duplicate of the h1 wording at h3, short/long/hidden/below-fold headings that must be skipped, an image and a button at known document coordinates, a hovered custom-cursor element, a dark section with a light heading and an outlined light-text button, and three bare-button controls. A second page is visited. Everything is read back through the real `GET_SCRAPS` message from the extension's scraps page: headings carry only the eight typographic keys with no background of any kind; the outlined button records `rgb(28, 32, 38)`; `role="button"` prose is absent while border-only and icon-only controls are present; positions match within 3px with the document scroll size attached. The scraps page is then driven: the headings filter shows five tiles and no backdrop nodes; the "Sign in" tile's patch decodes to median luminance 32.7. Finally the browser is closed and relaunched on the same profile: scraps persist and the cold service worker still collects on first load.

Screenshots are in the PR Evidence comment.


spencerc99:
## Preview evidence

**Scraps collects headings, scrap position, and button backdrops; skips bare text tagged as buttons.**

[Open preview](https://7e27e91a.we-were-online-website.pages.dev/installation/live/?clean=2&screen=clicks) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-mubmmvxb-81df23c4) · Commit `4fefec6`

### All kinds, archive pile

All kinds, archive pile

Headings beside other kinds. The light-text Sign in button is readable on its saved dark backdrop.

1200×850

### Headings filter

Headings filter

Five headings as bare text in their own color, no background patches. Pale ones are faint by design.

1200×850

### Buttons filter

Buttons filter

Only controls with a visible button look survive; role=button prose is absent.

1200×850

### After browser restart

After browser restart

Scraps persist across a relaunch; the cold service worker still collects.

1200×850

spencerc99 (commented):
## Code review

### Medium

- **Hidden headings under a transparent ancestor are still collected** (`extension/src/collectors/ScrapCollector.ts:608`). The check only reads the heading element's own computed `opacity`; opacity is not inherited, so `<div style="opacity: 0"><h1>private/alternate copy</h1></div>` keeps `opacity: 1` on the heading, has nonzero geometry, and can pass this IntersectionObserver. That violates the new “never actually seen” filter and stores text the reader could not see. Check effective visibility through the ancestor chain (including ancestor opacity) before emitting, and cover that case in the collector tests.

- **Modern translucent CSS colors are treated as opaque** (`extension/src/collectors/scrapUtils.ts:77`). `colorAlpha` only understands legacy comma-separated `rgba()`; computed values such as `rgb(0 0 0 / 0.2)`, `oklch(... / 0.2)`, or `color(display-p3 ... / 0.2)` either produce fewer than four split parts or miss the regex and return `1`. Buttons using those supported color syntaxes therefore lose the ancestor backdrop, so light text can become unreadable when reconstructed (and a fully transparent modern color can also make bare text look like a button). Parse alpha across the CSS Color 4 forms—or normalize through a browser color parser—and add translucent/transparent modern-syntax tests.

spencerc99:
Rebased onto current `origin/main` (picked up #507's source filters and inline search, #513, #461, #503, #502, #504, #505) and addressed both review findings.

## Hidden headings under a transparent ancestor — `7c550b35`

`opacity`, `display`, `clip` and `clip-path` do not inherit, so the old check on the heading's own computed style passed a heading inside an `opacity: 0` wrapper. Effective visibility now walks the ancestor chain, capped at 12 like the backdrop walk, and runs only after the cheaper text-length, size and per-page-cap checks.

`visibility` needs no walk: it inherits, and a descendant that sets `visibility: visible` genuinely shows through a hidden parent, so the element's own computed value is already correct.

**Do the other kinds share the gap?** Yes, and worse — buttons, images and SVG icons had no visibility check at all. All three now take the same walk, placed before the expensive work (SVG serialization, style picking). Cursors do not need it: they are found by hovering, which an invisible element cannot receive.

Covered by three collector tests (heading under faded/undisplayed/clipped wrappers, a heading faded four levels up, and one for button + image + icon). Verified they fail without the ancestor walk.

## Modern translucent CSS colors treated as opaque — `74b4e443`

`colorAlpha` now parses slash-alpha generically for every CSS color function (number or percentage), plus legacy four-component `rgba()`/`hsla()`, the `transparent` keyword, and `#rrggbbaa`/`#rgba` hex. Alpha is clamped to 0–1.

Per the no-silent-defaults rule, an unrecognized syntax returns a typed `ColorAlpha = number | undefined` rather than 1, and each caller decides what that means:
- `borderSideShows`: unreadable border color counts as paint, so a bordered control is kept.
- `isBareTextButton`: unreadable background counts as chrome, so a real button is not discarded as prose.
- `backdropFor`: unreadable own background records a backdrop rather than assuming none is needed.
- the backdrop walk: only a color read as fully solid stands in as the backdrop, so an unreadable ancestor is passed over instead of adopted.

Something that merely looks functional, such as an unresolved `var()`, is not treated as a color. Tests cover `rgb(0 0 0 / 0.2)`, `oklch(… / 0.2)`, `color(display-p3 … / 0%)`, `lab(… / 0.5)`, `hsl(… / 20%)`, legacy `hsla()`, the hex forms, clamping, and the unknown cases; `isBareTextButton` and the backdrop walk are re-verified against those forms.

## Also in this push

- `d21ef3ab` — the type filter moved into #507's filters popover, so the heading option and its test moved with it.
- `39fc171a` — typed the heading test helper to the heading member of `ScrapItem` and dropped two unused default React imports (one was a pre-existing `check:extension-website` error on `main`).

## Verification

`bun run test:extension` 1484 passed / 186 files, `bun run check:extension` clean, `bun run check:extension-website` clean (including the pre-existing error above), `bun run build-extension` succeeds.

Real-browser harness (headless Chromium, built `chrome-mv3`, read back through the real `GET_SCRAPS` message) re-run and extended with the two requested cases:

- **Heading inside an `opacity: 0` ancestor** — absent, alongside the existing hidden/too-short/too-long/below-fold cases. The scraps page shows "4 of 8" under the headings filter.
- **`background: rgb(0 0 0 / 0)` control on a dark section** — it carries a `1px` border, so the bare-text rule keeps it, and the harness asserts that: collected, own `backgroundColor` `rgba(0, 0, 0, 0)`, and `backdropColor` `rgb(28, 32, 38)` from the dark bar. Before the alpha fix this control would have read as opaque and recorded no backdrop.

Also covered: capture and dedup rules, allowlisted styles with no markup stored, document-space positions for image/button/heading/cursor, the scraps page rendering and heading filter, persistence across reload, and a cold start (browser shut down and relaunched on the same profile, then a page with wording never seen before collects on the freshly woken worker). Result `passed`, no page errors; screenshots in `/tmp/headings-evidence/`. The harness was run from `smoke-tests/` and deleted, not committed.

Two things worth flagging, both pre-existing in the harness rather than caused by these changes: it measured the document size on the wrong page (it had already navigated away), and its cold-start step re-visited an already-captured heading under a different query string, which storage correctly dedups by domain plus wording. Both are fixed in the harness.

Leaving the review open for you to resolve.

PR #509, opened 2026-09-21
partykit: coalesce presence attachment writes

spencerc99:
PresenceServer serialized every accepted message into the complete WebSocket attachment, even when its broadcast was coalesced or a keepalive repeated unchanged values. In the real local Workers runtime, a burst of 90 cursor messages produced 90 native attachment writes. This change produces two: the immediate broadcast and the trailing broadcast.

Accepted updates now coalesce by connection and channel at the existing 60 Hz broadcast boundary. Each changed attachment is persisted before its changes are published. Unchanged keepalives still reach peers but cause no attachment write. Joining clients flush pending state before receiving their snapshot; close/error discard pending writes, and closed sockets are skipped. A scheduled flush prevents hibernation until persistence completes.

The native attachment write remains the exact total-size guard, including PartyServer metadata and all channels. Cloudflare documents a 16,384-byte serialized limit. Rejected writes restore both the PartyServer cache and native attachment before returning an explicit client error; rejected candidates never reach peers. JSON byte counts would not measure this boundary correctly, and local workerd's `node:v8.serialize` is unimplemented, so this adds no approximate cap or duplicate sizing write. A rejected candidate batch retains its prior accepted snapshot.

Validation:

- `PRESENCE_BASE_REF=origin/main node --test partykit/__tests__/presenceRuntime.mjs`: reproduces the regression, failing with 90 writes instead of 2.
- `node --test partykit/__tests__/presenceRuntime.mjs`: passes with real PartyServer, native Workers WebSockets, and native attachment serialization. Covers burst coalescing, unchanged keepalive delivery with zero writes, aggregate overflow rollback, channel clear, actual hibernation confirmed by a different Durable Object instance, late join snapshots, early join flush/timer cancellation, and close/error cleanup. Expected error diagnostics are captured and asserted. The fixture observes native serialization without replacing its behavior.
- Focused presence tests: 24 passed. Common package tests: 18 passed. `bun run check:partykit` and `git diff --check`: passed.
- Full Worker dry-run bundle: passed (Wrangler emitted its existing experimental-secrets and unspecified-environment configuration warnings; no deployment).
- Full PartyKit suite: 304 passed with loopback HTTP permitted. The first sandboxed run failed its existing local HTTP test; that test and the full suite passed with network permission. Existing quarantine tests still print intentional `storage exploded` / `kv down` errors and Yjs client-ID warnings, so the broad suite output is not pristine under repository guidance.

This removes a demonstrated amplification path; it does not establish that this code caused the provider's platform-generated Durable Object storage resets. The incident evidence also includes ordinary PartyServer objects. Cloudflare still needs to investigate the object IDs, timestamps, and error references. No production deployment or production mutation was performed. Verification is local Worker runtime coverage, not a deployed production or visual-browser check.

Only Worker internals and tests change. No published package behavior/API, package dependencies/exports, starter templates, extension features/permissions, UI, or public documentation changes are involved, so no changeset, starter update, screenshot, or extension release note is required. Internal rationale is retained in the worktree's ignored `internal-docs/2026-09-21-presence-attachment-batching.md`.

References: [Cloudflare attachment API](https://developers.cloudflare.com/durable-objects/api/websockets/), [Durable Object lifecycle](https://developers.cloudflare.com/durable-objects/concepts/durable-object-lifecycle/), and installed PartyServer 0.5.5's attachment-cache and connection implementations.


spencerc99 (commented):
## Code review

PR #510, opened 2026-09-21
partykit: add presence-only local dev server

spencerc99:
## Summary

- add a dedicated local Worker entry and Wrangler config that export only `PresenceServer`
- add `bun run dev-server:presence` for presence UI and transport verification without Supabase or control-plane secrets
- add a real initialization smoke that removes the full Worker secrets, starts Wrangler, opens `/parties/presence/:room`, and asserts the initial `presence-sync` message
- document the local command and the existing multi-client presence smoke

The root cause was module evaluation, not request routing: the combined Worker entry imports `PartyServer`, which imports `db.ts` and constructs the Supabase client before Wrangler can route a request to `PresenceServer`.

## Architecture

The production entry and `partykit/wrangler.jsonc` remain unchanged. The full `PartyServer` path still constructs its persistence client eagerly and fails closed when required configuration is missing. The new local config exposes only the volatile `PresenceServer` Durable Object and has no `Main` binding, Supabase secrets, admin token, bridge secret, quarantine KV binding, routes, or deployment target.

This avoids making production persistence lazy or optional. Those alternatives would add a production state in which the persisted server can start without its authoritative storage dependency.

## Verification

- `./node_modules/.bin/tsc -p partykit`
- `bun test partykit/__tests__/presenceServer.test.ts` (6 passed)
- `bun test partykit/__tests__` with loopback access (301 passed, 0 failed)
- `PLAYHTML_PARTYKIT_SMOKE_PORT=1996 bun smoke:partykit:presence-local` (passed with all four full-Worker secrets removed)
- `PLAYHTML_PARTYKIT_PORT=1998 bun run dev-server:presence`, followed by the existing three-client presence cursor smoke at 30 Hz (90 sends, 186 receives, all clients observed both peers)
- presence-only Wrangler dry-run (72.88 KiB; only `env.Presence`)
- full Worker Wrangler dry-run (1,044.01 KiB; `env.Main`, `env.Presence`, and quarantine KV intact)
- `env -u SUPABASE_URL -u SUPABASE_KEY -u ADMIN_TOKEN -u PARTYKIT_BRIDGE_SECRET PLAYHTML_PARTYKIT_PORT=1997 bun run dev-server` still exits with `supabaseUrl is required`
- `git diff --check`

No deployment or production mutation was performed. No changeset, public docs update, or starter-template update is needed because published package behavior and public APIs are unchanged.


spencerc99:
CI blocker on current `main`: [#502](https://github.com/spencerc99/playhtml/pull/502) added nullable page-data tests that currently fail on the base branch. After rebasing this PR onto `b470ebd8`, I reproduced the same failure locally with:

```sh
bun run --cwd packages/playhtml vitest run src/__tests__/page-data.test.ts
```

Both `clears nullable object data` cases expect the listener updates to include `null`, but receive only the preceding `{ name: "Alice" }` value. This PR does not touch `packages/playhtml` or those tests, so I am keeping that core regression out of this PartyKit-only change.

PR-owned validation on rebased head `eb174ace` is green: PartyKit type-check, 301 PartyKit tests, focused presence tests, the secret-free real Wrangler/WebSocket smoke, and `git diff --check`.

spencerc99 (commented):
## Code review

PR #511, opened 2026-09-21
react: stabilize package metadata test paths

spencerc99:
## Summary

- Resolve React package metadata fixtures from the test file instead of the process working directory.
- Keep the package contract test runnable from the repository root and from `packages/react`.

## Rationale

The test previously read `package.json`, `vite.config.ts`, and source fixtures relative to `process.cwd()`. Running it from the repository root produced misleading missing-file and package-contract failures even though the same test passed from `packages/react`.

## Reproduction

Before this change, the repository-root invocation failed all three tests. The package-directory invocation passed after building package outputs.

## Verification

- `bunx vitest run packages/react/src/__tests__/package-metadata.test.ts --no-typecheck` from the repository root: 3 passed
- `bunx vitest run src/__tests__/package-metadata.test.ts --no-typecheck` from `packages/react`: 3 passed
- `bun run check:react`: passed
- `bun run test:react`: 56 unit tests and 3 integration tests passed

## Release notes

This changes test path resolution only. It does not change published package behavior, public APIs, documentation, or starter templates, so it does not need a changeset or documentation updates.


spencerc99 (commented):
## Code review

PR #512, opened 2026-09-21
Prevent extension checks from emitting JavaScript

spencerc99:
## Summary

- prepare WXT metadata explicitly during setup and before extension typechecks
- enforce non-emitting extension checks in both the TypeScript config and command
- make doctor point to the explicit WXT repair command
- typecheck the shipped Internet Scraps grid, pile, and preview routes

## Why

A fresh `bun run check:extension` without `extension/.wxt/tsconfig.json` reported the missing base config and then emitted 398 JavaScript files beside TypeScript sources across extension and package source trees. Those files can shadow the TypeScript sources on later runs. Depending only on the workspace postinstall hook also leaves setup vulnerable when WXT preparation does not complete after dependency installation.

The website check included only `src` and `internet-map`, so the built `scraps-grid`, `scraps-pile`, and `scraps-preview` routes could ship without TypeScript checking.

## Verification

- `bun run setup`
- `bun run test:doctor` (5 tests passed)
- `bun run doctor`
- `bun run check:extension`
- removed generated `.wxt` metadata, reran `bun run check:extension`, and confirmed WXT regenerated it with zero shadowing JavaScript files
- `bun run check:extension-website`; `tsc --listFilesOnly` includes all four scraps source files
- `bun run build-extension`
- `bun run build-extension:website`
- `git diff --check`

## Known blockers

`bun run test:extension` still fails the existing `image fingerprints > aborts a response that never finishes` case at its 15-second timeout. The full run passed 173 test files and 1,098 tests before that failure. The isolated test also times out. This branch does not touch the fingerprint implementation or test.

PR validation also fails two nullable-object notification assertions in `packages/playhtml/src/__tests__/page-data.test.ts`, introduced by current main in #502. The focused local run passes 15 tests and fails the same two assertions. This branch does not touch package source or tests.

## Follow-up

The Vite website build discovers additional route entrypoints that remain outside `extension/website/tsconfig.json`. Adding all of them at once exposed existing type errors in admin, components-preview, shared keypress/installation code, social-playground, and touches. This PR keeps strict checking and adds only the three confirmed Internet Scraps gaps. A follow-up cleanup should resolve those errors before expanding the typecheck to every Vite route.

This is internal verification tooling. It does not change published package behavior, extension permissions, or user-facing release notes.


spencerc99 (commented):
## Code review

PR #513, opened 2026-09-21
playhtml: notify nullable page-data listeners

spencerc99:
## Summary

- Observe the page-data parent key while an object or array root has a deep observer.
- Rebind the channel observer after local or remote root replacements.
- Cover nullable object and array notifications, including remote object → `null` → object transitions.

## Root cause

PR #502 correctly replaces the parent map entry when object or array page data becomes `null`. The channel was observing only the former child value, so Yjs persisted the replacement but did not notify listeners through that child's deep observer.

## Verification

- Before the fix: `bun run test:playhtml -- src/__tests__/page-data.test.ts` failed 2 of 17 tests because both nullable object clears omitted the `null` listener value.
- After the fix: `bun run test:playhtml -- src/__tests__/page-data.test.ts` passes 17 tests.
- `bun run test:playhtml` passes 52 files and 545 tests.
- `bun run check:playhtml` passes.
- `bun run build-packages` passes.

## Release impact

- No additional changeset. `.changeset/clear-nullable-page-data.md` already includes the subscriber notification behavior in the pending `playhtml` patch.
- No docs change. The page-data guide already documents nullable clears, and this fix makes the documented behavior observable through `onUpdate`.
- No starter change. The public API shape and usage are unchanged.


spencerc99 (commented):
## Code review

PR #514, opened 2026-09-21
extension: bound image fingerprint downloads

spencerc99:
## Summary

- Settle image fingerprint downloads when the existing 10-second deadline expires, even if an aborted response body stays pending.
- Keep abort-based request cancellation, response validation, byte limits, hashing, and concurrency behavior unchanged.

## Root cause

`fetchImageFingerprint()` treated `AbortController.abort()` as if it guaranteed the pending `reader.read()` would settle. The real local HTTP fixture showed that the abort timer fired after 10 seconds, but the body read could remain pending under the full parallel suite until teardown closed the connection. The function therefore had no independent completion deadline.

The download now races against the existing deadline. When the deadline wins, the function returns `undefined` and still aborts the request for cleanup.

## Verification

- Reproduced the original failure in the full extension suite: the abort timer fired, but the test timed out after 15 seconds.
- Ran the isolated real-server abort test three times: 3/3 passed at about 10 seconds each.
- Ran the complete image fingerprint test file: 6/6 passed.
- Ran `bun run test` from `extension/`: 174 files and 1,101 tests passed.
- Ran `bun run check:extension`.
- Ran `git diff --check`.

## Release checklist

- No changeset: this does not change a published package.
- No public docs or starter changes: the extension behavior has no public API.
- No screenshots: this is a background reliability fix with no user-facing surface.
- No `extension/PENDING.md` entry: this is a narrow reliability fix without a new user-visible workflow.


spencerc99:
CI note: `build-site` currently fails in the base branch on `packages/playhtml/src/__tests__/page-data.test.ts` because nullable object listeners do not emit the `null` update. This is reproducible from current `origin/main` and unrelated to this PR, which changes only `extension/src/storage/imageFingerprints.ts`. PR #513 is the exact base-branch fix and has a passing `build-site` check: https://github.com/spencerc99/playhtml/pull/513

PR #514 remains unstacked on `main`. Its local extension evidence is green: 174 test files and 1,101 tests passed, including the real-server timeout regression.

spencerc99 (commented):
## Code review

- **Medium — `extension/src/storage/imageFingerprints.ts:76`: timed-out downloads can escape the concurrency bound.** When the timeout wins this race, `fetchImageFingerprint()` resolves immediately, so `ImageFingerprints.fingerprint()` deletes the pending entry, decrements `active`, and starts queued work. However, `downloadImageFingerprint()` can still be blocked on `reader.read()`—the exact condition this change is meant to handle—leaving its response, reader, and network request alive. Repeated hanging image bodies can therefore accumulate beyond the intended two-request limit and eventually exhaust service-worker/network resources. Make the downloader itself abandon and cancel the reader on the deadline (or retain the slot until underlying cleanup is confirmed), and assert that the server connection closes after timeout rather than only that the outer promise resolves.

spencerc99:
Addressed the timeout/concurrency review on b0bba346 after rebasing onto current main.

The timeout no longer wins an outer promise race. It aborts the request, while the downloader owns the response reader, explicitly cancels it on abort, awaits cancellation, and releases the reader lock before returning. ImageFingerprints therefore keeps its concurrency slot until underlying body cleanup has settled.

Regression evidence:
- Real loopback test now asserts a never-finishing response closes after timeout.
- Stress regression sends four never-finishing image bodies through the two-slot pool, verifies peak server concurrency remains at two, and verifies all server connections are closed before processing returns.
- Focused fingerprint suite: 7/7 passed.
- Extension typecheck passed.
- Full extension suite: 186 files, 1,485 tests passed with pristine output.

PR #516, opened 2026-09-21
extension: wait for content test initialization

spencerc99:
## Summary

- Wait for content-script presence and collector initialization to finish before resetting shared test mocks.
- Keep the real feature-access path and bounded public-identity assertion in the regression test.
- Use the announcement check and migration storage read as observable completion signals for the two independent initialization branches.

## Root cause

`contentScript.main()` starts presence and collector initialization without awaiting either branch. The first test ended after two microtasks. Under load, its initialization continued after `beforeEach` reset the hoisted mocks for the next test. The second test could then receive the first test's nameless identity, while the collector migration read used the reset storage mock.

The runtime identity contract is correct. This change makes the test wait for the work it starts instead of changing production behavior or weakening the identity assertion.

## Verification

- Focused regression test passed in six concurrent runs. Before the fix, two of three concurrent runs failed with the missing-name mismatch.
- Related identity and lifecycle suite passed: 6 files, 11 tests.
- `bun run check:extension` passed.
- Full `bun run test:extension`: the content development feature test passed. The suite reported 1,100 passing tests and the existing `imageFingerprints > aborts a response that never finishes` timeout as its only failure.

No screenshot or `extension/PENDING.md` update is needed because this changes test synchronization only.


spencerc99 (commented):
## Code review

PR #517, opened 2026-09-22
extension: scrap collage builder (create mode on the scraps page)

spencerc99:
# Internet scraps: a create mode for making collages

## Summary

Adds a "create" mode to the internet scraps page. The scraps you have collected
while browsing become material: you drag them into a fixed frame, arrange and
trim them, and save the result as a collage you can come back to.

What is in it:

- **A studio.** A drawer of your scraps beside a fixed frame. Drag or click a
  scrap to place it, then move, resize, rotate, crop, mirror, reorder, duplicate
  or remove it. Rotate and scale also work as modal gestures (press `R` or `S`,
  move the pointer, click to confirm). Images can have their background cut
  away. Everything is undoable.
- **Four sizes and a paper.** Postcard, postcard tall, square and wide, each at
  a true pixel size the studio only ever zooms to fit. The paper takes a tone
  and, by default, the same soft grain the rest of the scraps page is printed
  on.
- **Saving that looks after itself.** A collage is written as soon as it has
  something in it and keeps itself up to date as you work; there is no save
  button, just a quiet word in the corner saying where things stand.
- **A history of what you have made.** Every collage with its picture, when it
  was made, and how many pieces are in it. Cards turn over to show the pages
  each piece came from. You can reopen one to keep editing, copy one to try a
  variation, or delete it.
- **Reading and reaching.** Hold `i` to see where every piece came from, tagged
  onto the piece itself. Click the same spot again to reach a piece buried under
  others, or right-click for a list of everything stacked there.
- **Export.** Any collage can be saved out as a PNG.

Everything is local to the browser. Nothing is uploaded.

## Why

The scraps page could only show you what you had collected, in an arrangement
the page chose. This lets you make something out of it — the editing half of the
"turn one browsing session into a piece of art" idea, with the self-composing
version to follow.

A collage stores links and layout plus one baked picture, never copies of the
images themselves, so a saved collage stays small and still shows where its
material came from even after the original scrap has been pruned away.

## Notes for review

- **Gated.** The whole mode sits behind the internal `SCRAP_COLLAGES` flag on
  top of `SCRAPS`. With the flag off the page is exactly as it was, and the mode
  switch is not rendered at all.
- **Migration.** `extension/worker/migrations/0004_scrap_collages_feature.sql`
  registers the feature at the `internal` stage. **It has to be applied before
  the flag can be granted to anyone server-side.** Until then the feature is
  reachable only through the local developer override.
- **Stored collages are real data.** Two narrow, named upgrades bring older
  saved collages forward: one for collages made before sizes and paper existed,
  one for collages made before paper could carry grain (those keep a plain
  paper, so nothing already saved changes how it looks). Both are covered by
  tests built from the exact earlier shapes, and both run once and write back.
- **No new permissions.** `wxt.config.ts` is untouched.
- **No package changes.** Nothing under `packages/` is touched, so no changeset.
- **No changelog entry.** The feature ships dark behind an internal flag; the
  bullet belongs in whichever PR opens it up to everyone.
- **Two shared files touched lightly.** `ScrapCollage.tsx` exports its content
  renderer, its heading font sizing and its collage styles so the studio draws
  scraps exactly as the browse view does, and `LocalEventStore` gained the
  import path. Both are additive.
- **Headings and button backdrops, from #508.** Headings are a full kind in the
  tool: they filter under their own chip, place at the size the shared renderer
  draws them, crop and flip like anything else, and bake through the same
  foreignObject path buttons take, so the page's own typography survives into
  the PNG. A button's recorded backdrop is now painted in the bake as well as on
  screen, which is what keeps light text read off a dark section legible once it
  is sitting on pale paper. `position` needs nothing here and got nothing.

## Verification

**Automated, in a real browser.** `smoke-tests/scrap-collage.mjs` drives the
built extension (`chrome-mv3-dev`) in an isolated headless Chromium with its own
profile, against local loopback pages only, with all other network blocked. Run
with `bun run smoke:scrap-collage`.

The flow starts where a person would: it browses three fixture pages so the
collectors gather one of every scrap kind (picture, button, icon, heading,
cursor), opens the scraps page, switches to create, and works from there. It asserts
against IndexedDB and the decoded PNG, not against the screen.

What it covers:

- **Every kind bakes.** All five scrap kinds are placed, spread out and saved;
  the stored PNG is decoded and sampled under each piece to prove it actually
  drew, and that the paper's grain is really in the picture.
- **A backdrop bakes behind its text.** A light-text outlined button, collected
  off a dark section, bakes with dark pixels behind its words while the paper
  around it stays pale.
- **Saving.** A brand-new empty collage writes nothing at all; placing a piece
  stores it with nothing pressed; a move survives a reload immediately after;
  leaving flushes what is pending; `Cmd+S` writes at once.
- **Saving when the pictures cannot be fetched.** With the image host blocked,
  the arrangement still saves, the previous picture is kept rather than
  overwritten, and the status says the preview is out of date. Export, which is
  asked for out loud, still fails out loud. A first-ever bake that fails stores
  the arrangement with an explicit "no preview" and the history draws a quiet
  face.
- **The chrome.** The piece strip appears above the selected piece, stays on
  screen for a piece at the top edge, hides during a drag, and carries no
  to-front/to-back buttons; the paper popover opens, changes paper and grain,
  and closes on Escape; the bottom bar holds nothing but export and done.
- **Reading sources.** Holding `i` tags every piece and outlines it, no two tags
  overlap, the hovered piece gets the fuller tag, the tags never take the
  pointer or change the selection, they vanish on release and on window blur,
  and a peek changes neither the stored record nor the baked picture.
- **Reaching a buried piece.** Three pieces stacked on one spot: clicking walks
  top → middle → bottom → top, dragging still moves the piece in hand, the
  stacking order never changes, the hover outline marks the right piece, and the
  right-click list shows all three and selects the one picked.
- **The card's back.** A card turns over, its lines match the record's own
  provenance, and only http(s) pages are linked.
- **Copying.** Duplicating a card produces a second collage with its own id and
  the same geometry; editing the copy leaves the original untouched, down to its
  last-changed time.
- **The drawer at real scale.** Against a real 5,000-scrap export: thumbnails
  keep their own proportions to within a pixel, three equal columns, and the
  number actually mounted stays bounded at every size (70 / 55 / 47). A
  thirty-piece collage built from that material bakes to a real PNG.
- **Transparency backing.** A genuine JPEG gets no chequer behind it; a genuine
  half-transparent PNG does.

The real-export section seeds through the extension's own `IMPORT_EVENTS`
message rather than any page control, so it exercises the real import path. The
run loads the production build, because the development build fetches its
modules from the WXT dev server that the run blocks.

The section that needs a real export is skipped with a clear log line unless
`SCRAPS_EXPORT` points at one, so the test runs anywhere. It needs no image
tooling and no network.

**Also run, all green:**

- `bun run lint` (every package) — clean
- `bun run test:extension` — 201 files, 1800 tests passed
- `bun run check:extension-website` — clean
- `bun run build-extension` (production) — 4.31 MB
- `node smoke-tests/scrap-collage.mjs` with `SCRAPS_EXPORT` set — passed, 7
  collages, real-export section run

## Screenshots

From `/tmp/collage-evidence/`:

- `08-studio-1280.png` — the studio at 1280 wide
- `05-piece-strip.png` — the tools floating beside the selected piece
- `07-paper-popover.png` — sizes, paper tones and grain
- `12-peek-held.png` — holding `i` over a mixed collage
- `09-pieces-here.png` — the list of everything under the pointer
- `13-card-front.png` / `14-card-back.png` — a collage card, both faces
- `16-real-export-tray.png` — the drawer against the real 5,000-scrap export
- `11-baked-preview.png` — a baked collage PNG


spencerc99:
## Preview evidence

**Create mode on the scraps page: arrange collected scraps into collages, with headings and button backdrops from #508 carried through.**

[Open preview](https://7e27e91a.we-were-online-website.pages.dev/installation/live/?clean=2&screen=clicks) · [Review all evidence](https://pr-evidence.spencerc99.workers.dev/review/ev-muccszmv-7c2755a2) · Commit `4996855`

### The studio

The studio

Drawer of collected scraps beside a fixed frame at 1280 wide, with the piece tools and the bottom bar.

1500×980

### Piece tools

Piece tools

The selected piece's tools float just above its box, at a constant on-screen size rather than the frame's zoom.

1500×980

### Sizes and paper

Sizes and paper

The popover holding the four collage sizes, the paper tones, and whether the paper carries its grain.

1500×980

### Holding i to read sources

Holding i to read sources

Every piece tagged with where it came from; the tags never overlap, take the pointer, or change the selection.

1500×980

### Reaching a buried piece

Reaching a buried piece

Right-clicking a pile lists everything stacked under the pointer so a buried piece can be picked directly.

1500×980

### The drawer at real scale

The drawer at real scale

Against a real 5,000-scrap export: three columns at each scrap's own proportions, only what is in view mounted.

1500×980

### A collage card, front

A collage card, front

The history card: the collage's picture, when it was made, and how many pieces are in it.

1500×980

### A collage card, back

A collage card, back

Turned over to the pages each piece came from; only http(s) pages are linked.

1500×980

### A baked collage

A baked collage

The stored PNG for a collage holding all five scrap kinds, printed on the paper's own grain.

1500×980

spencerc99 (commented):
## Code review

### Medium — Reopening a collage drops its last good preview before the replacement is known to work

`extension/src/entrypoints/scraps/useCollageAutosave.ts:89` initializes `previewRef` (and, at line 96, `storedRef`) empty even when `startsStored` is true. On the first edit after “keep editing,” the schedule deliberately skips the pre-write bake, so line 139 stores an undrawn preview before starting the replacement bake. If that bake fails—or the page closes after the arrangement write—the valid preview loaded with the collage has already been permanently erased, contrary to the promised fallback behavior. Pass the loaded record into the hook and seed both refs from it so arrangement writes retain the existing preview until a new bake succeeds.

### Medium — Superseded bakes can overwrite a newer preview

`extension/src/entrypoints/scraps/useCollageAutosave.ts:182` persists every completed bake before the state machine gets a chance to reject its token. If revision 1’s bake is slow, revision 2 can write and finish its own bake first; when revision 1 finally resolves, it reads the current `storedRef` and overwrites revision 2’s preview with the stale image. The later `bakeSucceeded` event is ignored, but the IndexedDB write has already happened. Check that `stateRef.current.bakingToken === token` immediately after the await and before mutating either preview ref or storing the result.

spencerc99:
Both review findings are fixed, each in its own commit with a unit test that fails before and passes after. I did not resolve the review threads.

**1. Reopening dropped the last good preview — `9a134c93`**

`useCollageAutosave` now takes a `reopening` record and seeds both `previewRef` and `storedRef` from it, so the first arrangement write after "keep editing" carries the picture the collage was loaded with until a new bake succeeds. `CollageStudio` passes the `editing` record it already holds.

Confirmed against the real thing before fixing — the smoke test reproduced the data loss exactly: `drawn: true, previewBytes: 13698903` became `drawn: false, previewBytes: 0`. A 13 MB preview was being permanently destroyed by one edit with the image host away. After the fix the bytes are identical across the edit while the arrangement still advances.

Covered by two unit tests (loaded preview retained on the first write; retained when the re-bake fails) and a new smoke section that reopens a saved collage, blocks the image host, moves a piece, waits for the write, and asserts the stored preview is byte-identical.

**2. Superseded bakes overwrote a newer preview — `2989bcf0`**

The bake now checks `stateRef.current.bakingToken === token` immediately after the await, before touching either preview ref or storing. The reducer already rejected stale tokens; the side effects simply ran before reaching it.

I also guarded the **failure** path, which had the same flaw in the other direction: a superseded bake's error would overwrite `previewProblemRef` and mark a perfectly healthy newer picture as "preview out of date". Both paths now return early on a stale token.

The unit test drives two overlapping bakes with deferred promises, lets revision two land first, then resolves revision one and asserts its image never reaches the store.

**Sibling checked, no change needed:** the export path (`download` in `CollageStudio.tsx`) bakes to a local object URL for the download and never touches `previewRef`, `storedRef`, or `store`, so it cannot lose or overwrite a stored preview. Its failure is surfaced out loud, which is the intended difference from a save.

**Verification, all green:**

- `bun run lint` — clean
- `bun run test:extension` — 202 files, 1803 tests passed (1800 + 3 new)
- `bun run check:extension-website` — clean
- `bun run build-extension` — 4.31 MB
- `node smoke-tests/scrap-collage.mjs` with `SCRAPS_EXPORT` — passed, 7 collages, real-export section run

